PR #10, opened 2026-05-14
Fix header merge to prevent duplicate keys in lite TLS proxy

mpenn:
## Description

`_forward_headers` could land the inbound `Content-Length` and the proxy's
own `Content-Length` in the outgoing dict under different cases (e.g.
lowercase `content-length` from the OpenAI Python SDK plus capitalized
`Content-Length` re-added by the proxy). Python's `http.client` doesn't
deduplicate, so both went out on the wire. nginx in front of
`integrate.api.nvidia.com` rejects duplicate `Content-Length` as an
RFC 7230 §3.3.3 violation with a bare-HTML `400 Bad Request`.

Symptom: agent traffic returned `200` on `GET /v1/models` but `400` on every
`POST /v1/chat/completions`. Small direct curl bodies slipped through
nginx's lenient parsing path; ~40 KB agent payloads did not.

## Changes

- `_forward_headers` canonicalizes keys to lowercase on insertion. The three
  proxy overrides (`host`, `content-length`, `connection`) at the bottom of
  the function naturally win over any inbound casing — no skip list, no
  maintenance trap if a future contributor adds a fourth override.
- Inline comment captures the *why* (RFC 7230 §3.3.3, nginx-400) so the
  lowercasing isn't "cleaned up" by a later refactor.
- New `scripts/test_host_tls_proxy.py` — 6 stdlib `unittest` cases,
  self-runnable as `python3 scripts/test_host_tls_proxy.py`. Tests assert
  via case-insensitive lookups so the suite remains valid if the
  implementation later switches to e.g. `wsgiref.headers.Headers`.


pastorsj (approved):
lgtm

PR #17, opened 2026-05-24
Phoenix LLM spans now show with real prompts and completions

mpenn:
## Summary

End-to-end NeMo-Relay observability for Hermes via a **single persistent sidecar gateway**. One daemon serves every Hermes surface — PID-1 gateway, Slack/Outlook bridge-driven turns, and interactive `hermes chat` TUI — producing per-turn ATIF trajectory files at `/tmp/atif/*.json` plus Phoenix LLM spans carrying the real prompt + completion, paired tool spans (no orphans), and per-turn agent root spans. All four surfaces land in the same correlated Phoenix session.

## Architecture

`start.sh` launches a long-running NeMo-Relay daemon at sandbox startup (`start_nemo_relay_sidecar` helper, listens on `127.0.0.1:4040`). Every Hermes process discovers it via `NEMO_RELAY_GATEWAY_URL` in env — exported explicitly on PID-1's launch list and via `_PROXY_ENV_FILE` (sourced by `/sandbox/.bashrc`) for interactive shells. The daemon fails hard on `/healthz` timeout: dumps the last 30 lines of `/tmp/nemo-relay.log` to stderr and `exit 1`s. Silent telemetry loss is worse than a noisy startup error.

Two event paths feed the daemon:

- **Shell hooks** (`generate-config.ts`): `on_session_*`, `pre/post_llm_call`, `subagent_stop` via `nemo-relay hook-forward hermes`.
- **In-process plugin** (`plugins/nemo-relay/`): `pre/post_api_request` + `pre/post_tool_call` — forwards the rich kwargs (full request messages, response SDK object, stable tool_call_ids) that the shell path strips.

`nemo-relay-finalize-hook` runs as a second `on_session_end` command to synthesize a per-turn `on_session_finalize` (Hermes's native finalize only fires on idle expiry). Daemon dispatches to ATIF (always-on) and OpenInference (gated by `PHOENIX_COLLECTOR_ENDPOINT`).

Interactive `hermes` invocations from sandbox shells resolve through a PATH-prepended CLI shim at `/usr/local/lib/nemoclaw/bin/hermes` — warns on `setup`/`doctor` about the in-sandbox config lifecycle and `exec`s the upstream binary for everything else. Telemetry flows because the shell's `NEMO_RELAY_GATEWAY_URL` propagates to the exec'd process; the in-process plugin POSTs to the sidecar like any other Hermes surface.

## Sandbox workarounds

[...]

PR #18, opened 2026-05-27
Tidy up Hermes NeMo-Relay integration

mpenn:
## Summary

Follow-up to #17. No behavior change in normal use — removes dead code,
inlines a small adjacent module, streamlines docs, and adds upstream-tracking
markers.

## Changes

- **Remove `_export_otel_bsp_tunings`** from `start.sh`. The
  `OTEL_BSP_MAX_EXPORT_BATCH_SIZE=1` / `SCHEDULE_DELAY` /
  `EXPORT_TIMEOUT` env vars were a workaround for the OpenShell L7 proxy
  dropping batched OTLP POSTs in the nemo-flow-0.2.0 era. Validated by
  commenting them out and confirming telemetry still flows under
  0.3.0-beta.2 with the sidecar architecture.

- **Drop `_cap_payload` + `_MAX_PAYLOAD_BYTES`** from
  `plugins/nemo-relay/__init__.py`. Defensive payload truncation with no
  upstream precedent — every other NeMo-Relay integration (claude-code,
  codex, cursor, openclaw) sends payloads without size caps. Sidecar's
  axum body limit handles oversize payloads at the right layer if it
  ever matters.

- **Inline `nemoclaw_patches.py` into `sitecustomize.py`** (Slack
  catch-all slash-command handler). One less PYTHONPATH bootstrap file.

- **Remove the Discord-facade machinery** from `sitecustomize.py`
  (`NEMOCLAW_DISCORD_FACADE_URL` + ~80 lines of URL-rewriting code).
  Unused — no documented use case, and upstream NemoClaw's tests
  explicitly assert this env var is NOT set. The Slack-egress
  placeholder rewriting (the actually load-bearing function) stays.

- **Trim the module docstring** in `plugins/nemo-relay/__init__.py` from
  ~28 to ~10 lines. Inline workaround comments at each call site stay
  where they're useful.

- **Add `TODO(upstream)` markers** in `plugins/nemo-relay/__init__.py`
  and `nemo-relay/finalize-hook`. Documents that the in-process plugin
  is retirable when [NousResearch/hermes-agent#29724](https://github.com/NousResearch/hermes-agent/pull/29724)
  lands a built-in NeMo-Flow telemetry plugin, and that the
  finalize-hook is retirable when NeMo-Relay exposes a per-turn ATIF
  flush knob.

- **README**: brief ATIF lifecycle note — files accumulate, sandbox
  teardown clears them, `find -type f -mtime +N -delete` for manual
  prune.

Net: −176 / +18 lines across 7 files (1 deleted).

## Test plan

- [ ] Rebuild image; sandbox comes up clean.
- [ ] `pgrep -af nemo-relay` shows one daemon; `curl /healthz` returns 200.
- [ ] Drive one Slack/Outlook/TUI turn each — ATIF file lands per turn,
      Phoenix spans correlate to the same sidecar PID. Same behavior as
      before this PR.


pastorsj (approved):
Testing on all platforms, Outlook + Slack. LGTM

PR #21, opened 2026-05-27
Add configurable Phoenix project name for trace routing

mpenn:
## Summary

Adds `PHOENIX_PROJECT_NAME` as a configurable knob so users running multiple
deployments of this example against the same Phoenix instance can keep their
traces separated under named projects. Without this, every example's spans
fall back to `service.name` (`hermes-agent`) and route to the default bucket.

## How it works

Phoenix routes spans to a named project by reading the `openinference.project.name`
OpenTelemetry resource attribute. NeMo-Relay's `OtlpSectionConfig` exposes a
`resource_attributes` HashMap that attaches verbatim to the OTLP resource —
this PR templates a single entry into that block and plumbs the value through
the existing Dockerfile ARG → ENV → `@@TOKEN@@` sed-substitution pattern used
for `PHOENIX_COLLECTOR_ENDPOINT`.

## Changes (~23 lines across 4 files)

- **`agents/hermes/nemo-relay/plugins.toml.in`**: new
  `[components.config.openinference.resource_attributes]` table with
  `"openinference.project.name" = "@@PHOENIX_PROJECT_NAME@@"`. File header
  comment updated to list all three substitution tokens (was naming only
  `PHOENIX_COLLECTOR_ENDPOINT`).
- **`agents/hermes/Dockerfile`**: new `ARG PHOENIX_PROJECT_NAME="default"` /
  matching `ENV` near the existing `PHOENIX_COLLECTOR_ENDPOINT` block; sed
  expression extended with `-e "s|@@PHOENIX_PROJECT_NAME@@|...|g"`.
- **`.env.example`**: documents the override knob alongside the existing
  Phoenix block.
- **`README.md`**: one row added to the env-vars table.

## Defaults and overrides

- **Default**: `default` — the literal string, which is Phoenix's catch-all
  bucket name. Users who don't override see no behavior change from before
  this PR (traces still arrive; they just nominally land under `default`
  rather than under `service.name`).
- **Override**: set `PHOENIX_PROJECT_NAME=<name>` in `.env`, rebuild the
  sandbox. Useful for multiple-deployment workflows or A/B comparisons.

## Test plan

- [ ] `bash scripts/tear-down.sh && bash scripts/bring-up.sh` builds clean.
- [ ] `openshell sandbox exec --name <sandbox> -- cat /etc/nemo-relay/plugins.toml`
      shows the `[components.config.openinference.resource_attributes]` table
      with the project name substituted.
- [ ] Drive one Hermes turn with `PHOENIX_COLLECTOR_ENDPOINT` set; the
      resulting span appears under the project named `default` (or the
      override) in Phoenix's UI.
- [ ] Set `PHOENIX_PROJECT_NAME=test-override` in `.env`, rebuild, drive one
      turn — Phoenix project becomes `test-override`.
- [ ] Each span's `service.name` attribute still reads `hermes-agent`
      (separate from project routing).

### DIFF
--- a/examples/personal-community-sentiment-triage/.env.example
+++ b/examples/personal-community-sentiment-triage/.env.example
@@ -109,6 +109,12 @@ GITHUB_READONLY_REPO=NVIDIA/OpenShell
 # sandbox at it through the OpenShell host bridge.
 # PHOENIX_COLLECTOR_ENDPOINT=http://host.openshell.internal:6006/v1/traces
 
+# Phoenix project name. Sets openinference.project.name on every exported
+# span so Phoenix routes traces to a named project (vs. falling back to
+# service.name). Defaults to "community-sentiment-triage" if unset; override
+# to keep multiple deployments separate in the same Phoenix instance.
+# PHOENIX_PROJECT_NAME=personal-community-sentiment-triage
+
 # ── Host-services security (TOKEN_CACHE_SALT is optional) ────────────────
 # Salt for encrypting the MS Graph token manager's MSAL session cache.
 # The compose default (in extras/docker-compose.yml) is shared across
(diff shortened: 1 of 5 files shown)


pastorsj (approved):
LGTM

PR #24, opened 2026-05-28
Consolidate README demo catalog

pastorsj:
## Summary

- Replace the directory-focused README table with a combined demo catalog
- Include local demos plus selected external demos from `brevdev/nemoclaw-demos`
- Make Getting Started generic now that multiple demos have their own setup guides

## Verification

- Verified local guide paths exist
- Verified external demo guide links return HTTP 200
- `markdownlint` was not installed locally


pastorsj:
Closing so this can be reopened from the contributor fork instead of a branch on the upstream repository.

PR #25, opened 2026-05-28
Consolidate README example catalog

pastorsj:
## Summary

- Replace the directory-focused README table with a combined demo catalog
- Include local demos plus selected external demos from `brevdev/nemoclaw-demos`
- Make Getting Started generic now that multiple demos have their own setup guides

## Verification

- Verified local guide paths exist
- Verified external demo guide links return HTTP 200


mpenn (approved):
Looks great, approve.

PR #35, opened 2026-06-08
fix: truncate large hook payloads before forwarding to NeMo Relay

### LINKED ISSUE #22: Large Hermes hook payloads are rejected before Phoenix export
## Scope
`examples/personal-community-sentiment-triage` on current `main`.

## Current main behavior
The example Hermes plugin forwards the full hook payload to NeMo Relay. It does not truncate or compact the payload before sending it to `/hooks/hermes`.

Current `main` builds `nemo-relay-cli` version `0.3.0-beta.2`. In that version, `/hooks/hermes` is handled with Axum `Json<Value>`:

```rust
.route("/hooks/hermes", post(hermes_hook))

async fn hermes_hook(..., Json(payload): Json<Value>)
```

The NeMo Relay router does not set `DefaultBodyLimit::max(...)` or `DefaultBodyLimit::disable()` for hook routes. Axum therefore applies its default JSON extractor body limit, which is about 2 MiB. A Hermes hook payload larger than that is rejected before `hermes::adapt(...)` runs. There is no automatic truncation.

This is separate from the NeMo Relay gateway `/v1/chat/completions` path, which explicitly buffers request bodies with `MAX_BODY_BYTES = 100 * 1024 * 1024`.

## Reproduction
1. Check out current `main`.
2. Configure the example normally with Phoenix enabled and bring it up:
   `cd examples/personal-community-sentiment-triage`
   `bash scripts/tear-down.sh`
   `bash scripts/bring-up.sh`
3. From inside the running sandbox, post a synthetic Hermes hook payload larger than 2 MiB to the relay hook endpoint:

```bash
openshell sandbox exec --name hermes-direct -- python3 - <<'PY'
import json
import os
import urllib.error
import urllib.request

base = os.environ.get("NEMO_RELAY_GATEWAY_URL", "http://127.0.0.1:4040")
url = base.rstrip("/") + "/hooks/hermes"
payload = {
    "hook_event_name": "pre_api_request",
    "request": {
        "body": {
            "model": "test-model",
            "messages": [
                {"role": "user", "content": "x" * 2_200_000}
            ],
        }
    },
}
data = json.dumps(payload).encode("utf-8")
req = urllib.request.Request(
    url,
    data=data,
    headers={"Content-Type": "application/json"},
    method="POST",
)
try:
    with urllib.request.urlopen(req, timeout=10) as response:
        print(response.status)
        print(response.read(500).decode("utf-8", "replace"))
except urllib.error.HTTPError as error:
    print(error.code)
    print(error.read(500).decode("utf-8", "replace"))
PY
```

4. Repeat the same command with the content length below 2 MiB, for example `"x" * 1_500_000`.
5. Inspect Phoenix and `/tmp/atif` for the corresponding hook/span records.

## Observed
- Payloads above about 2 MiB return HTTP `413 Payload Too Large` from the `/hooks/hermes` endpoint.
- The Hermes adapter does not run for that oversized hook payload.
- No Phoenix/ATIF record is produced for that rejected hook payload.
- The example plugin on `main` does not surface the non-2xx response from `client.post(...)`, so the Slack turn can appear to work while observability silently misses the hook/span.

## Why this matters
Large but otherwise valid LLM turns can exceed the hook endpoint limit. In that case, the agent interaction may complete, but the example's Phoenix/ATIF telemetry no longer reflects the actual turn.

nanookclaw:
## Summary

The NeMo Relay Axum server applies a default ~2 MiB body limit on `/hooks/hermes`. Large but valid LLM turns silently return HTTP 413, so the interaction completes but Phoenix/ATIF telemetry is missing for that turn.

## Root Cause

`_forward()` sends the full serialized hook payload without checking size. When a Hermes turn produces a large payload (multi-message conversations, verbose tool outputs, long reasoning chains), the Axum `Json<Value>` extractor rejects it before `hermes::adapt(...)` runs.

## Change Made

**File:** `examples/personal-community-sentiment-triage/agents/hermes/plugins/nemo-relay/__init__.py`

Added a `_truncate_payload()` step in `_forward()` that progressively shrinks oversized payloads while preserving correlation metadata (`task_id`, `session_id`, `model`, etc.) and hook event structure:

- **Phase 1:** Truncate message content strings with a proportional per-message budget
- **Phase 2:** Truncate `response.raw_response` and `assistant_message` content
- **Phase 3:** Truncate tool `args` and `result` strings
- **Phase 4:** Aggressive — replace all content with `[truncated]` marker, set `_truncated: true`

Threshold is 1.5 MiB, safely under Axum's ~2 MiB limit. Small payloads pass through unchanged (identity return, zero overhead).

This is the client-side complement to any future server-side `DefaultBodyLimit` increase in NeMo Relay itself.

```
.../plugins/nemo-relay/__init__.py   | 146 ++++++++++++++++++++
 1 file changed, 146 insertions(+)
```

## Testing

- Verified small payloads pass through unchanged (identity return)
- Verified 2 MiB message content is truncated below 1.5 MiB
- Verified 10-message conversations get proportional truncation
- Verified correlation metadata (task_id, session_id, model) is preserved through all phases
- Verified response and tool call payloads are handled
- Python syntax validation passes

Fixes #22

### DIFF
--- a/examples/personal-community-sentiment-triage/agents/hermes/plugins/nemo-relay/__init__.py
+++ b/examples/personal-community-sentiment-triage/agents/hermes/plugins/nemo-relay/__init__.py
@@ -207,6 +207,151 @@ def _serialize_response_object(response: Any) -> Optional[dict]:
 
 
 # ---------------------------------------------------------------------------
+# Payload truncation
+# ---------------------------------------------------------------------------
+
+# Axum's default Json<Value> extractor rejects bodies above ~2 MiB.
+# Stay safely under that limit so large-but-valid LLM turns still
+# produce telemetry.
+_MAX_PAYLOAD_BYTES = 1_500_000  # 1.5 MiB — headroom under Axum's ~2 MiB limit
+
+
+def _truncate_str(text: str, max_len: int = 4096) -> str:
+    """Truncate a string, appending a marker so consumers know it was cut."""
+    if len(text) <= max_len:
+        return text
+    return text[:max_len] + "...[truncated]"
+
+
+def _estimate_size(payload: dict) -> int:
+    """Fast serialized-size estimate (UTF-8 bytes of JSON)."""
+    try:
+        return len(json.dumps(payload, default=str, ensure_ascii=False).encode("utf-8"))
+    except Exception:
+        return 0
+
+
+def _truncate_payload(payload: dict) -> dict:
+    """Iteratively shrink a hook payload until it fits under the Axum body limit.
+
+    Truncation order (largest fields first):
+      1. message content strings in request.body.messages
+      2. response.raw_response deep content
+      3. response.assistant_message content
+      4. tool args / result strings
+
+    Correlation metadata (task_id, session_id, model, etc.) is never touched.
+    """
+    size = _estimate_size(payload)
+    if size <= _MAX_PAYLOAD_BYTES:
+        return payload
+
+    # Work on a mutable copy
+    payload = dict(payload)
+
+    # --- Phase 1: truncate LLM message content ---
+    request_body = (payload.get("request") or {}).get("body")
+    if isinstance(request_body, dict):
+        messages = request_body.get("messages")
+        if isinstance(messages, list):
+            budget_per_msg = max(512, _MAX_PAYLOAD_BYTES // max(len(messages), 1) // 4)
+            changed = False
+            for i, msg in enumerate(messages):
+                if isinstance(msg, dict):
+                    content = msg.get("content")
+                    if isinstance(content, str) and len(content) > budget_per_msg:
+                        messages[i] = {**msg, "content": _truncate_str(content, budget_per_msg)}
+                        changed = True
+                    elif isinstance(content, list):
+                        # multimodal content blocks
+                        for j, block in enumerate(content):
+                            if isinstance(block, dict):
+                                text = block.get("text")
+                                if isinstance(text, str) and len(text) > budget_per_msg:
+                                    content[j] = {**block, "text": _truncate_str(text, budget_per_msg)}
+                                    changed = True
+            if changed:
+                payload["request"] = {**payload.get("request", {}), "body": {**request_body, "messages": messages}}
+
+        size = _estimate_size(payload)
+        if size <= _MAX_PAYLOAD_BYTES:
+            return payload
+
+    # --- Phase 2: truncate response content ---
+    response = payload.get("response")
+    if isinstance(response, dict):
+        raw = response.get("raw_response")
+        if isinstance(raw, dict):
+            # Truncate choice text content
+            for key in ("choices", "output", "content"):
+                val = raw.get(key)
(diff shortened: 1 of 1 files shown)


abhisawa-Nvidia:
@nanookclaw would you file a GH issue on [NeMo Relay](https://github.com/NVIDIA/NeMo-Relay/issues) repo please? We can handle the case in Relay directly.

nanookclaw:
Filed NVIDIA/NeMo-Relay#255 with the Relay-side oversized hook payload handling context.

WilliamK112:
The Relay-side follow-up requested above looks done now: NVIDIA/NeMo-Relay#261 closed NVIDIA/NeMo-Relay#255 and shipped in NeMo Relay 0.4.0.

That release raises the default hook payload limit to 20 MiB, adds `NEMO_RELAY_MAX_HOOK_PAYLOAD_BYTES`, and includes regression coverage for hook payloads above Axum's default limit plus structured 413 responses. This example currently pins `nemo-relay-cli@0.3.0`, so the smallest follow-up in this repo may be to bump the pinned CLI to `0.4.0` instead of maintaining plugin-side truncation here.

I prepared and locally validated a minimal two-file patch for that path (`agents/hermes/Dockerfile` version/comment update plus one stale `0.3.0` comment cleanup in `nemo-relay/finalize-hook`). Validation passed:

- `python scripts/check_license_headers.py --check`
- `git diff --check`
- `bash -n examples/personal-community-sentiment-triage/scripts/*.sh`
- `python -m py_compile` over the example Python files

Happy to send that minimal bump if maintainers prefer this route.

apurvvkumaria:
Maintainer hold at exact head `b79b10132b243d15b5f4fabc7597a9a747bc62c8`: a dependency-owner decision is required before technical work resumes.

The Relay-side payload-limit fix requested in this thread shipped in NeMo Relay 0.4.0, while this repository still pins 0.3.0. Please choose the supported path: upgrade Relay and rely on its configurable server limit, or retain a client-side truncation policy with an explicitly accepted data-loss contract. If the upgrade path is selected, land a minimal signed replacement, preserve @nanookclaw's attribution, and prove an oversized Hermes hook reaches Relay/Phoenix before closing #22 and this PR. Do not close this contribution until that replacement is merged and verified.

apurvvkumaria:
Closing this implementation following the maintainer decision to use the upstream server-side fix instead of client-side payload truncation.

NVIDIA/NeMo-Relay#261 resolved NVIDIA/NeMo-Relay#255 and shipped in NeMo Relay 0.4.0. That release raises the default hook payload limit, makes the limit configurable, and returns structured errors for oversized requests. Truncating the payload in this repository could silently discard telemetry and create a separate data-loss policy.

Any update to the Relay version pinned by this example should be handled as a focused dependency change with end-to-end validation. This closure supersedes the earlier hold that asked for that replacement to merge first.

Thank you @nanookclaw for identifying the payload-limit problem and contributing the proposed mitigation.

PR #43, opened 2026-07-02
Add Harness Engineering Playground (hep) example

mpenn:
## Summary
- Adds `hep`, a standalone CLI dev-tool example (not an OpenShell-run agent
  blueprint like the other examples) that automatically tunes a LangChain
  `deepagents` `HarnessProfile` against its own eval suite via a
  propose/apply/verify/roll-back loop.
- Built around two extension seams: `HarnessAdapter` (target-framework eval
  runner + config I/O — `DeepAgentsAdapter` is the first implementation) and
  `Optimizer` (search technique — `RalphLoopOptimizer` proposes fixes via a
  bounded, tool-using deep-agent session with real `read_file`/`edit_file`/
  `grep`/`glob` access to the SDK and eval-suite source).
- Bundles `deepagents` as a pinned git submodule (`external/deepagents`) as
  the default eval target, so the example works out of the box without a
  separate checkout.

## Test plan
- [x] `git submodule update --init && uv sync --extra deepagents`
- [x] `uv run pytest tests/unit_tests/ -v` — 27 passed
- [x] Followed `docs/verify-functionality.md` for an end-to-end live run
      against the bundled submodule


README.md
```diff
@@ -21,6 +21,7 @@ Some examples are included in this repository. Others currently live in [brevdev
 | ---- | ----------- | ---- |
 | Personal Community Sentiment Triage | Pairs a Hermes harness with an OpenShell sandbox and community-signal integrations across Slack, Outlook, live read-only GitHub REST, GitHub discussion mirrors, and NVIDIA forum mirrors. | [Guide](examples/personal-community-sentiment-triage/README.md) |
 | Hermes Brev Launchable | Provides a notebook path from a fresh Brev CPU instance to a working NemoClaw-managed Hermes sandbox, including installation, onboarding, API verification, and terminal access. | [Guide](examples/hermes-launchable/README.md) |
+| Harness Engineering Playground | A CLI dev tool (not an OpenShell blueprint) for automated, eval-driven harness profile improvement, with pluggable target-framework adapters and optimization techniques. | [Guide](examples/harness-engineering-playground/README.md) |
```
slopp, line 24: We really need to overhaul the top-level README, this is a list of brevdev demos not the examples in this repo

examples/harness-engineering-playground/.env.example
```diff
@@ -0,0 +1,20 @@
+# Frontier model used to propose fixes. RALPH_API_KEY takes precedence over
+# ANTHROPIC_API_KEY so you can use a different key/endpoint for the improvement
+# loop than for the model under evaluation.
+RALPH_API_KEY=
+ANTHROPIC_API_KEY=
```
slopp, line 6: This feels a little confusing given the model under test is Nemotron Ultra, with an openai prefix? How do users wire through the base url and key for the model target, eg if they wanted to use build or baseten or some other open model?

slopp (approved):

PR #50, opened 2026-07-13
Add Payment Operations Hermes example

aadesoba-nv:
## Description

Add `examples/payment-ops-hermes`, a regulated payment-operations blueprint built around a constrained Hermes agent, NeMo Relay evidence, and a platform-enforced maker/checker boundary.

FinGuard can screen synthetic ISO 20022-shaped payments, explain holds, and prepare evidence for human review. OpenShell deliberately provides no payment-rail egress, so an agent release attempt is denied independently of the prompt or model. A separately attributed host-side human checker is the only component that can call the mock rail.

The example includes:

- a custom Hermes + NeMo Relay sandbox image
- real Hermes model/tool hook forwarding to Relay
- local ATIF trajectories and Phoenix OpenInference export
- a curated three-record public OFAC fixture with deterministic positive and negative controls
- an idempotent phased lifecycle for initial bring-up and Brev resume
- separate agent and human-checker telemetry
- FSI production-adaptation guidance

The root examples table and `THIRD-PARTY-NOTICES` are updated for the new example. The curated source snapshot is derived from standalone source commit `aadesoba-nv/payment-ops-hermes@361b49e` without carrying that repository's history into this contribution.

## Verification

- [x] `python scripts/check_license_headers.py --check`
- [x] `git diff --check`
- [x] Shell syntax checks for all example lifecycle scripts and entrypoints
- [x] Python compile checks for all example Python sources and the Relay finalize hook
- [x] JavaScript syntax check for the FinGuard UI
- [x] Six of six payment-control fixture scenarios pass
- [x] Curated OFAC positive control holds `ACH-2003` on `BANK ROSSIYA`
- [x] Curated OFAC negative control clears `WIRE-1007` for human review
- [x] Markdown local-link validation
- [x] Private-environment and generated-artifact scans
- [x] Standalone source deployment verified on Brev with Hermes, NeMo Relay, ATIF, Phoenix, sandbox denial, and human checker evidence
- [x] Fresh end-to-end bring-up from `examples/payment-ops-hermes` in this community branch

## Release And Compliance

- [x] No secrets, local `.env` files, private certificates, snapshots, or token caches are included.
- [x] Third-party dependency changes are reflected in `THIRD-PARTY-NOTICES`.
- [x] Public documentation is free of internal-only links or private workspace details.
- [x] Commits include DCO sign-off (`git commit -s`).


examples/payment-ops-hermes/policy.yaml
```diff
@@ -0,0 +1,54 @@
...
+  phoenix:
+    name: phoenix
+    endpoints:
+      - host: host.openshell.internal
+        port: 6006
+        protocol: rest
+        enforcement: enforce
+        allowed_ips:
+          - 10.0.0.0/8
+          - 172.0.0.0/8
```
nv-nedelman-1: this one seems pretty broad/including some public IPs. I think it should be `172.0.0.0/12`. @aadesoba-nv is there a specific reason for this broader range?

examples/payment-ops-hermes/policy.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - 192.168.0.0/16
+          - fc00::/7
+        rules:
+          - allow: { method: POST, path: "/**" }
+    binaries:
+      - { path: /usr/local/bin/nemo-relay }
+  ofac:
```
nv-nedelman-1: as far as I can tell, the agent running inside openshell doesn't actually call out to this API endpoint. if it's truly not needed/used, we should remove it

examples/payment-ops-hermes/policy.yaml
aadesoba-nv: Good catch. I checked this against the pinned OpenShell v0.0.53 policy documentation. Because host.openshell.internal is an exact user-declared hostname, it may resolve to an RFC1918 address without an allowed_ips override. I’ll remove the broad IP allowlist entirely rather than replace it with another subnet.

examples/payment-ops-hermes/policy.yaml
aadesoba-nv: Agreed. The sandbox therefore does not need OFAC network egress, so I removed that policy block.

examples/payment-ops-hermes/docs/demo-script.md
nv-nedelman-1, line 1: @aadesoba-nv I think this file should be removed - since it specifically refers to showing this demo at a booth/workshop.

examples/payment-ops-hermes/scripts/_lib.sh
```diff
@@ -0,0 +1,59 @@
...
+  [[ -f "${pid_file}" ]] || return 0
+  pid="$(<"${pid_file}")"
+  if kill -0 "${pid}" 2>/dev/null; then
+    echo "Stopping ${name} (PID ${pid})"
+    kill "${pid}"
+  fi
+  rm -f "${pid_file}"
+}
+
+sandbox_phase() {
```
nv-nedelman-1, line 46: `openshell sandbox list` returns the `PHASE` with ascii color characters, which [breaks the case statement](https://github.com/NVIDIA/nemoclaw-community/pull/50/changes#diff-f91f37ae48d08dfab804547f7e5279749f89eb56bf8d7080ecb64db141c35c60R44) that maps to `ready`. I did not encounter this when running `scripts/bring-up.sh` the first time, only in subsequent invocations. 

claude gave me this change to parse the actual `ready` string and it works well:
```
sandbox_phase() {
  local name="${1:-$NEMOCLAW_SANDBOX_NAME}"
  openshell sandbox list 2>/dev/null | awk -v wanted="$name" '
    { gsub(/\033\[[0-9;]*m/, "") }
    NR > 1 && $1 == wanted { print $NF; found = 1; exit }
    END { if (!found) print "Missing" }
  '
}
```

nv-nedelman-1:
Hi @aadesoba-nv I made two more minor comments. I'll get this merged once addressed, thank you!

examples/payment-ops-hermes/scripts/_lib.sh
aadesoba-nv: done

examples/payment-ops-hermes/docs/demo-script.md
aadesoba-nv: done

PR #100, opened 2026-08-04
ci(dco): check sign-offs before push

apurvvkumaria:
## Summary

- add a repository pre-push hook that rejects branch updates containing commits without valid DCO trailers
- use one shared DCO range checker locally and in GitHub Actions
- document the one-time local hook installation command
- add focused coverage for accepted and rejected pushes

## Why

The required GitHub DCO check currently detects missing trailers only after a commit reaches a PR branch. The local hook gives contributors the same result before Git sends the branch update while preserving the required GitHub check as the authoritative merge gate.

Local hooks remain user-controlled and can be bypassed. This change does not claim to enforce policy inside user-owned forks without installation.

## Verification

- `bash -n scripts/check_dco_signoffs.sh .githooks/pre-push`
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` — 8 tests passed
- `python3 scripts/check_license_headers.py --check` — 253 files passed
- `python3 scripts/check_label_taxonomy.py --check` — 64 labels valid
- `git diff --check`

ShellCheck was not available locally; Bash syntax validation passed.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+remote_name="${1:-origin}"
+repo_root="$(git rev-parse --show-toplevel)"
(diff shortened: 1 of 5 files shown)


apurvvkumaria:
Closing in favor of the NemoClaw-aligned DCO model: one Signed-off-by declaration in each pull request description instead of a local pre-push hook and per-commit trailers. A focused replacement PR will follow.

PR #52, opened 2026-07-14
feat(examples): add blackwall-guard partner recipe (BlueTier)

bluetieroperations-create:
## What this adds

`examples/blackwall-guard/` — a community integration that adds a **pre-action risk gate** to an OpenClaw agent running under NemoClaw. It hooks `before_tool_call` and screens every tool call through the external [BLACK_WALL](https://blackwalltier.com) `forecast()` service **before it runs** — returning **GO / CAUTION / STOP**, a risk score, named red flags, and an Ed25519-signed, offline-verifiable receipt.

- **observe** mode (default): logs the verdict, never blocks — safe to trial.
- **enforce** mode: **STOP** blocks; **CAUTION** blocks by default (this runtime has no interactive approval surface), carrying the red-flag detail in the block reason. Configurable via `cautionAction`.
- Defense-in-depth against destructive commands, irreversible writes, exfiltration, or fund movement a compromised / prompt-injected agent might attempt — independent of the model's own judgment.

## Contents

`index.ts` (plugin), `openclaw.plugin.json` (manifest), `skills/blackwall-policy` + `skills/blackwall-verify`, and `index.test.ts` — a Vitest suite covering the `before_tool_call` decision state machine, the HTTPS-only credential guard, and the proxy CONNECT-header size cap.

## Context

This began as a `nemoclaw-blueprint` plugin PR on the main repo (NVIDIA/NemoClaw#4692). Per the maintainer's guidance there — that an independent third-party runtime integration belongs in `nemoclaw-community` rather than core — I've moved it here, as I offered on that PR. The code is unchanged from the version reviewed there (including the security fixes from that review). The README frames it clearly as an independent third-party integration calling an external service, not a NemoClaw-supported surface.

Signed-off-by: bluetieroperations-create <270213615+bluetieroperations-create@users.noreply.github.com>
## Checks

- **DCO**: commit is signed off.
- **License Headers**: `scripts/check_license_headers.py --check` passes locally (`.ts` files carry SPDX headers).
- No new dependencies; nothing added to `THIRD-PARTY-NOTICES`.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

PR #105, opened 2026-08-06
feat(examples): add x402-payment-gate partner recipe (BlueTier)

bluetieroperations-create:
## Related Issue

Companion to #52 (blackwall-guard) — same contributor, complementary scope. No issue to close.

## Description

Adds a partner recipe at `examples/recipes/partners/bluetier/x402-payment-gate/`: a maker/checker payment **boundary** for [x402](https://www.x402.org/) machine payments, in the pattern of `recipes/nvidia/payment-ops-hermes`. Before any x402 payment is signed, a host-side gate outside the sandbox asks a Blackwall verdict service to score the claim — counterparty reputation on Base, price-anomaly detection, OFAC sanctions screening, Sybil/graph signals — pre-signature.

- **Scope — a boundary demonstration, not a full agent runtime.** `bring-up.sh` stands up the boundary (rail + gate + policy-scoped sandbox); `verify.sh` exercises the in-sandbox maker path at the **intent-submission level** (`POST /v1/intents` over the scoped `host.openshell.internal` route → live verdict) plus the denied edge. It does **not** start an interactive Hermes agent — the skill under `agents/hermes/skills/` documents how a Hermes agent *would* drive this boundary; running that agent needs a full NemoClaw Relay+Hermes runtime and is a separate operator step.
- **Keyless, minimal data sharing.** Default endpoint is the free public instance; no API key exists anywhere. Only the payment claim (`counterparty, amount, asset, chain, resource`) leaves the sandbox — never tool payloads, never keys.
- **Maker/checker boundary.** The sandboxed agent (MAKER) can only submit intents; a host-side release gate (CHECKER) runs the mandatory verdict and is the only thing that signs and settles. The rail has **no** sandbox route — that denied edge is the boundary. No plugin ships here; the OpenClaw plugin lives in the Blackwall repo, contributed separately in #52.
- **Two authorization paths.** Automated release settles only on a fresh **GO**; human-approved release is a separate path for a HELD intent (named human + a host-only approval token on `127.0.0.1:8791`) that re-screens with a fresh verdict and still refuses a fresh **STOP**. Both decide before any signature exists; STOP is terminal on both.
- **Split listeners, discovered bind.** The sandbox reaches the host over the OpenShell bridge (`host.openshell.internal` → the `openshell-docker` gateway), not loopback, so `bring-up.sh` **discovers that bridge interface and binds the submit listener there** — reachable by the sandbox and the host, not the LAN, never `0.0.0.0`. Rail and approve stay on loopback.
- **Complete lifecycle.** `scripts/bring-up.sh` (`openshell sandbox create --from <recipe-root>` — the Dockerfile lives at the recipe root so its build context matches its `COPY` paths, `.dockerignore` trims it — and **always rebuilds the sandbox from the current checkout**, never reusing a stale one), `scripts/verify.sh` (live canaries + the in-sandbox denied-edge test, which **positively requires HTTP `403 policy_denied`** and fails on transport errors), `scripts/tear-down.sh`.
- **Enforcement architecture.** The mandatory decision runs in the host-side release gate (the `payment-ops-hermes` pattern). Wiring the same verdict into OpenShell Supervisor middleware is the ideal fail-closed tightening and the documented next step — offered for explicit maintainer agreement, not silently substituted.
- Single verified commit on current `main`, registered in the Partner Recipes catalog table. Note: #52 also adds a BlueTier row to that table — whichever PR lands second needs a one-line rebase there.

## Verification

- [x] `python3 scripts/check_license_headers.py --check`
- [ ] `python3 scripts/check_label_taxonomy.py --check` when governance metadata changes (n/a — no governance metadata changed)
- [x] `git diff --check`
- [x] Relevant example setup or syntax checks (`bash -n` on all lifecycle scripts; host unit suite 27/27; **validated end to end on a live OpenShell + Docker host** — bring-up → verify → tear-down, the in-sandbox maker path returns a fresh verdict and the rail is refused as HTTP 403 `policy_denied`)

## Release And Compliance

- [x] No secrets, local `.env` files, private certificates, snapshots, or token caches are included.
- [x] Third-party dependency changes are reflected in `THIRD-PARTY-NOTICES` (none added — stdlib Python only).
- [x] Public documentation is free of internal-only links or private workspace details.

Signed-off-by: Samuel Trujillo <bluetier.operations@gmail.com>

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/examples/recipes/partners/bluetier/x402-payment-gate/README.md
+++ b/examples/recipes/partners/bluetier/x402-payment-gate/README.md
@@ -1,0 +1,237 @@
+# x402-payment-gate: Verdict-Gated Payment Release
+
+A maker/checker payment boundary for [x402](https://www.x402.org/) machine
+payments, in the pattern of the
+[Payment Operations Hermes Assistant](../../../nvidia/payment-ops-hermes/README.md):
+the sandboxed Hermes agent can screen payments and **submit payment intents**,
+but it cannot sign or settle anything — it holds no key and has no network
+route to any payment rail. The **mandatory** decision lives in a host-side
+**release gate outside the agent sandbox**, which asks
+[Blackwall](https://blackwall-free.onrender.com/.well-known/x402) for a
+**GO / HOLD / STOP** verdict — counterparty reputation on Base, price-anomaly
+detection, OFAC sanctions screening, Sybil/graph signals — and only a GO is
+signed and settled. The decision is **genuinely pre-signature**: no signature
+exists until after the verdict, and the signing step is on a code path only
+the release branch reaches (pinned by unit test).
+
+The threat this closes: an injected or manipulated agent with payment
+capability is one signature away from paying a drainer, a sanctioned address,
+or a 100x-gouged invoice. Here a prompt cannot become a payment: the most a
+compromised agent can do is *submit an intent*, and the checker it cannot
+touch re-screens every intent before any money moves.
+
+> **Third-party integration — requirements & support.** This is an independent
+> community example contributed by BlueTier Operations, **not** a supported
+> part of NemoClaw core. It calls the external Blackwall verdict service — the
+> default public instance is **keyless and free** (no API key, no signup);
+> self-hosting is documented for real workloads. **Support** for this example
+> and the service is provided by BlueTier Operations, not NVIDIA — contact
+> <bluetier.operations@gmail.com>.
+
+> **Scope — a payment-*boundary* demonstration, not a full agent runtime.**
+> `bring-up.sh` stands up the boundary (mock rail + release gate + a
+> policy-scoped sandbox), and `verify.sh` exercises the in-sandbox **maker path
+> at the intent-submission level** (`POST /v1/intents` over the scoped
+> `host.openshell.internal` route → a live verdict) together with the denied
+> edge. It does **not** start an interactive Hermes agent. The skill under
+> `agents/hermes/skills/` documents how a Hermes agent *would* drive this
+> boundary; actually running that agent needs a full NemoClaw Relay+Hermes
+> runtime (inference provider, `nemoclaw-start`) and is a **separate operator
+> step, out of scope here**. What this recipe proves — on a real OpenShell host
+> — is the security property: a prompt in the sandbox cannot reach the rail, and
+> every intent is re-screened by the host-side checker before any money moves.
+
+## Architecture
+
+```mermaid
+flowchart LR
+    blackwall["External\nBlackwall verdict service"]
+    seller["External\nx402 seller (402 challenge)"]
+
+    subgraph host["Host machine"]
+        direction TB
+        gate["Release gate — CHECKER\n127.0.0.1:8790\nmandatory verdict → sign → settle"]
+        rail["Mock payment rail\n127.0.0.1:8780 · HOST ONLY"]
+        human["Named human operator\napproves HELD intents"]
+
+        subgraph supervisor["OpenShell sandbox supervisor"]
+            policy["L7 egress policy"]
+            subgraph sandbox["x402-payment-gate sandbox"]
+                agent["Hermes agent — MAKER"]
+                skill["blackwall-payment-gate skill\n+ stdlib client (advisory)"]
+                agent --> skill
+            end
+        end
+
+        skill -->|"advisory forecast"| policy
+        skill -->|"submit intent\nhost.openshell.internal:8790"| policy
+        policy --> gate
+        gate -->|"mandatory forecast"| blackwall
+        gate -->|"GO only: sign → settle"| rail
+        human -->|"approve HELD"| gate
+    end
+
+    policy -->|"advisory verdicts"| blackwall
+    seller -.->|"HTTP 402"| agent
+    sandbox -. "NO ROUTE — denied edge" .-> rail
+
+    style host fill:#f7f6ef,stroke:#8a8068,stroke-width:2px
+    style supervisor fill:#e7f0ff,stroke:#2b5fab,stroke-width:3px
+    style sandbox fill:#d8e8ff,stroke:#2b5fab,stroke-dasharray:5 3
+    style rail fill:#fce5cd,stroke:#c0392b,stroke-width:2px
+```
+
+The denied edge is the central security property: the rail route is absent
+from `policy.yaml`, so a prompt cannot override it. Note the deliberate bind
+asymmetry. The gate runs **two listeners**: a **submit/status** listener the
+sandbox reaches (via the `host.openshell.internal` route) and a **host-only
+approve** listener bound to `127.0.0.1:8791`. The sandbox reaches the host over
+the OpenShell bridge (`host.openshell.internal` → the `openshell-docker` network
+gateway), *not* host loopback — so the submit listener must bind that **specific
+bridge interface**: loopback would be unreachable from the sandbox (the maker
+path dies), and `0.0.0.0` would expose submission on every host interface (the
+LAN included). `bring-up.sh` discovers the bridge gateway and binds the submit
+listener there automatically (`RELEASE_GATE_BIND`; it defaults to a safe
+`127.0.0.1` when set by hand, and is never `0.0.0.0`). The **rail** binds
+`127.0.0.1` and is unreachable from the sandbox at the network layer — that
+asymmetry *is* the denied edge. Splitting
+the listeners means exposing the submit interface never exposes `/approve`: the
+sandbox can reach only the least-privileged operation (submit), while the
+human-only operations (approve, settle) live behind loopback. Everything
+in-sandbox is **advisory** — the skill's pre-check warns the user early, but
+nothing the agent does or skips changes what the gate enforces.
+
+## Key invariants
+
+- The agent can screen, explain, and submit intents; it cannot sign, release,
+  or settle a payment. No signing key exists inside the sandbox (in this demo
+  no real key exists anywhere — the rail signature is an explicitly labeled
+  simulation).
+- The sandbox has no route to the rail or any facilitator, even when a tool
+  or prompt attempts one. Its egress is the inference endpoints, the verdict
+  service, and the intent-submission route — nothing else.
+- **Only the host-side gate settles, and it does so through exactly two
+  authorization paths — never any other way.** Both run host-side and both
+  decide before any signature exists (verdict → sign → settle, enforced by
+  code and pinned by unit test):
+  1. **Automated release** — an intent settles unattended *only* on a fresh
+     **GO** verdict. HOLD and STOP never settle on this path.
+  2. **Human-approved release** — a **HELD** intent (a fresh HOLD, or a
+     verdict-service failure that held it) can be released by a **named human**
+     plus the approval token printed only to the gate's host-side log — a value
+     the sandbox cannot read; the held→releasing transition is lock-guarded
+     against concurrent double-approval. This path **re-screens with a fresh
+     verdict**: the named human overrides a HOLD, but a fresh **STOP** (e.g. the
+     payee became sanctioned between submit and approval) refuses the release
+     even with a valid operator and token — the re-forecast precedes signing, so
+     the decision stays pre-signature.
+  A STOP is terminal on both paths: no verdict, human, or token releases it.
+- A verdict-service failure HOLDS — the mandatory layer never fails open.
+- Only the payment claim (`counterparty, amount, asset, chain, resource`)
+  leaves the sandbox or the host — never tool payloads, never keys.
+
+## The verdict
+
+The gate (and the advisory skill) call `POST /v1/forecast-payment` with the
+claim and receive: `verdict` (GO/HOLD/STOP), `hard_stop`, `score`,
+`reasons[]`, per-signal breakdown, confidence, and a `receipt_id` +
+`report_token` for closing the loop after settlement
+(`POST /v1/report-outcome`). Gate mapping (shared with Blackwall's other
+guard integrations): **GO → release, HOLD → held for a human, STOP →
+refused**; anything unrecognized holds.
+
+## Lifecycle: bring-up, verify, tear-down
+
+```bash
+scripts/bring-up.sh     # host services (rail + gate) → reproducible sandbox
+                        #   image (pinned Hermes base + baked skill) → sandbox
+                        #   with policy.yaml applied whole
+scripts/verify.sh       # unit tests + FRESH payment canaries through the gate
+                        #   + denied-edge test from inside the sandbox
+scripts/tear-down.sh    # sandbox → host services
+```
+
+Only python3 is required to exercise the host boundary; docker + openshell
+add the sandbox phases. The policy is a **complete** sandbox policy: the
+inference routes mirror the
+[chief-of-staff recipe](../../../nvidia/developer-community-chief-of-staff/README.md)
+unbroadened, plus this recipe's verdict and intent routes — every in-sandbox
+HTTP caller is python3 or curl, exactly matching the binary allowlists.
+
+### Verification is live and fresh
+
+`verify.sh` initiates **new payments on every run** — nothing is inferred
+from pre-existing logs:
+
+1. Unit tests pin the decision core, including the verdict-then-sign order.
+2. Three fresh canaries go through the gate against the live verdict service:
+   a warm payee at its fair price (→ `released`, producing a fresh settlement),
+   a sanctioned payee (→ `refused`, never signed), an unknown payee
+   (→ `held`). The mock-rail ledger must grow by exactly the released one.
+3. From inside the sandbox: the intent route must work and the rail route
+   must be denied by the supervisor.
+
+A held intent can then be released only by a named human holding the
+approval token that the gate prints to its host-side log at startup — a
+value nothing inside the sandbox can read. The approve endpoint lives on the
+**host-only** listener (`127.0.0.1:8791`), which the sandbox has no route to:
+
+```bash
+curl -X POST 127.0.0.1:8791/v1/intents/<id>/approve \
+  -H 'X-Operator: <your name>' -H "X-Approve-Token: <from .run/gate.log>"
+```
+
+The held→releasing→released transition is lock-guarded, so two concurrent
+approvals can never both settle, and intent submission is bounded
+(`RELEASE_GATE_MAX_INTENTS`, default 1000) so a compromised agent cannot grow
+the store or hammer the verdict service without limit.
+
+## Files
+
+| Path | Purpose |
+|---|---|
+| `host/release_gate.py` | The checker: intent API + mandatory verdict → sign → settle. Stdlib only. |
+| `host/mock_rail.py` | Host-only settlement target with an auditable ledger; the denied edge. |
+| `host/test_release_gate.py` | Unit tests for the decision core (verdict mapping, order invariant, human-approval rules). |
+| `agents/hermes/skills/blackwall-payment-gate/SKILL.md` | The maker skill: advisory pre-check, intent submission, hold explanations. |
+| `scripts/blackwall_client.py` | Stdlib advisory client (`should_sign`: GO→sign, STOP→refuse, else escalate). |
+| `scripts/demo_verdicts.py` | Standalone four-scenario live verdict walkthrough. |
+| `policy.yaml` | Complete sandbox policy: inference + advisory verdict + intent routes; **no rail route**. |
+| `Dockerfile` | Reproducible sandbox image: pinned Hermes base + baked skill. Built with the recipe root as context (`--from <recipe-root>`), so its `COPY` paths resolve; `.dockerignore` trims that context. |
+| `scripts/bring-up.sh` · `verify.sh` · `tear-down.sh` | Lifecycle (above). |
+
+## Production notes and limits
+
+- **Replace the simulation at the edges, keep the boundary.** The mock rail
+  and simulated signature are demo stand-ins; production adopters wire the
+  release gate to a real wallet/signing provider so the same verdict gates
+  real signing — see the wallet and OpenClaw adapters in the Blackwall
+  repository (`integrations/wallets/`, `integrations/openclaw/`). For
+  OpenClaw-runtime agents, the related in-sandbox defense-in-depth hook is
+  contributed separately in the
+  [BLACK_WALL Preflight Guardrail](https://github.com/NVIDIA/nemoclaw-community/pull/52)
+  example.
+- **Enforcement architecture: host-side gate now, Supervisor middleware as the
+  documented next step.** An earlier review asked for the mandatory decision to
+  run inside OpenShell's `HTTP_REQUEST`/`PRE_CREDENTIALS` **Supervisor
+  middleware** (the decision function against in-flight egress, fail-closed,
+  inside OpenShell's own enforcement layer). This recipe instead places the
+  mandatory decision in a **host-side release gate outside the sandbox** — the
+  established pattern of the
+  [payment-ops-hermes recipe](../../../nvidia/payment-ops-hermes/README.md) — with
+  the rail denied to the sandbox by policy. That host-side boundary is what this
+  recipe validates end to end. Wiring the same verdict into Supervisor
+  middleware is the **ideal fail-closed tightening and the documented next
+  step**, once (and where) that configuration surface is available to community
+  recipes. Adopting the host-side gate as the accepted architecture for *this*
+  recipe is offered for explicit maintainer agreement; the middleware remains
+  tracked as follow-up rather than silently substituted.
+- **Verdict quality is advisory input, not an oracle.** Signals that depend
+  on seller-controlled inputs (e.g. the resource URL driving the category
+  price baseline) are HOLD-only and evadable by a motivated seller; the free
+  public instance is a shared demo tier (ephemeral state, idle spin-down,
+  ~60s cold start, no SLA). Self-host for real workloads.
+- **This example moves no real money.** Payments settle against the bundled
+  mock rail with simulated signatures. Extending to a real facilitator and a
+  funded wallet is a deliberate operator decision far outside this example's
+  scope.
--- a/examples/recipes/partners/bluetier/x402-payment-gate/host/release_gate.py
+++ b/examples/recipes/partners/bluetier/x402-payment-gate/host/release_gate.py
@@ -1,0 +1,426 @@
+#!/usr/bin/env python3
+
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Host-side release gate — the CHECKER in the maker/checker boundary.
+
+Runs OUTSIDE the agent sandbox (mirroring the payment-ops-hermes pattern):
+the sandboxed agent can only SUBMIT a payment intent here; it holds no
+signing capability and no route to the payment rail. This gate runs the
+mandatory Blackwall verdict, and a payment reaches sign-then-settle through
+exactly TWO host-side authorization paths, never any other way:
+
+  1. Automated release (process_intent): an intent settles unattended ONLY on
+     a fresh GO verdict. HOLD and STOP never settle here.
+  2. Human-approved release (process_approval): a HELD intent (a fresh HOLD,
+     or a verdict-service failure that held it) can be released by a NAMED
+     human plus the host-side approval token. This path re-screens with a
+     FRESH verdict first: the human overrides a HOLD, but a fresh STOP refuses
+     the release even with a valid operator + token.
+
+Both paths decide BEFORE any signature exists (the signing step lives on a
+code path only the RELEASE branch reaches, pinned by test_release_gate.py),
+so the decision is genuinely pre-signature. A STOP is terminal on both paths.
+
+Verdict mapping (family contract shared with the Blackwall langchain,
+wallet, and openclaw guards): GO -> release, HOLD -> hold for a human,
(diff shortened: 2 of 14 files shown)


examples/recipes/partners/bluetier/x402-payment-gate/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+Be precise about what the plugin is and is not. It cannot be skipped by
+*prompting* — unlike the skill, no injected instruction can talk the agent out
+of a hook the runtime fires on every tool call. But it runs **inside the agent
+sandbox**, so it is **defense-in-depth, not an independent security boundary**:
+a sufficiently capable compromised agent could in principle disable it. The
+layers the agent *cannot* touch are the OpenShell network policy (which caps
+where the sandbox can reach regardless of what the agent runs) and — because
+this integration is keyless — there is no credential in the sandbox to steal
+at all. Moving the forecast call into OpenShell's Supervisor middleware (an
```
apurvvkumaria: Please make Supervisor middleware the enforcement boundary in this recipe, not a future evolution. The current hook runs in the agent sandbox, can be disabled by a compromised agent, and only recognizes selected tool/parameter shapes; `handleBeforeToolCall` explicitly passes unrecognized calls through. Current OpenShell supports an outside-the-sandbox `HTTP_REQUEST/PRE_CREDENTIALS` middleware hook with fail-closed policy. Coordinate with OpenShell as needed and include that path here. If the current NemoClaw-managed lifecycle cannot register it yet, scope this as a non-enforcing demonstration and do not present the plugin as a mandatory payment guard.

examples/recipes/partners/bluetier/x402-payment-gate/plugin/index.ts
```diff
@@ -0,0 +1,542 @@
...
+  if (accept && typeof accept === "object") {
+    counterparty = counterparty ?? normalizeAddress(firstOf(accept, COUNTERPARTY_KEYS));
+    amount = amount ?? atomicToDecimal(accept.maxAmountRequired, USDC_DECIMALS);
+    asset = asset ?? firstOf(accept, ASSET_KEYS);
+    chain = chain ?? firstOf(accept, CHAIN_KEYS);
+    resource = resource ?? firstOf(accept, RESOURCE_KEYS);
+  }
+
+  // Signed X-PAYMENT header: decode for claim fields, and ALWAYS pass the raw
+  // header through for the server-side payload-sim cross-check.
```
apurvvkumaria: This sends the complete signed `X-PAYMENT` authorization, including its signature, to the external service. That is no longer a pre-signature check and it contradicts the repeated claim that only `{counterparty, amount, asset, chain}` leaves the sandbox and that signed payloads are never sent. Please base the default decision on the unsigned claim before signing. If a signed-payload simulation remains as an explicit optional feature, document exactly what leaves the sandbox and its trust/replay/retention implications instead of describing it as claim-only data sharing.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/bring-up.sh
```diff
@@ -0,0 +1,53 @@
...
+# `openshell ... --help` is authoritative if flags have moved.
+set -euo pipefail
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+EXAMPLE_DIR="$(dirname "$DIR")"
+SANDBOX_NAME="${SANDBOX_NAME:-x402-gate-demo}"
+
+echo "== 1/3 preflight =="
+command -v openshell >/dev/null || { echo "openshell not in PATH" >&2; exit 1; }
+[[ -n "${SANDBOX_IMAGE:-}" ]] || { echo "SANDBOX_IMAGE is required (OpenClaw-capable image with plugin/ baked in — see README)" >&2; exit 1; }
```
apurvvkumaria: Please make the example independently reproducible. It currently requires an unspecified custom image with the plugin already copied, installed, configured, and enabled, but the PR provides no Dockerfile, pinned base image, exact plugin destination, or build command. Add the build/install/enable artifacts and have bring-up create or select the resulting image so a contributor can follow the repository instructions from a clean checkout.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/verify.sh
```diff
@@ -0,0 +1,57 @@
...
+else
+  echo "skipped: npm or plugin/package.json not available (run from the repo checkout)"
+fi
+
+echo
+echo "== 3/3 interception (host-side, needs SANDBOX_NAME) =="
+if [ -z "${SANDBOX_NAME:-}" ] || ! command -v openshell >/dev/null 2>&1; then
+  echo "skipped: set SANDBOX_NAME and run where the openshell CLI is available."
+else
+  if openshell sandbox logs "$SANDBOX_NAME" 2>/dev/null | grep -E "\[blackwall-x402\] (observe|enforce)" | tail -3 | grep -q .; then
```
apurvvkumaria: This does not prove the current deployment intercepted a real payment call: any old matching log line makes the check pass, and the script never triggers the call it claims to verify. Please have verification issue a deterministic, uniquely identifiable payment-shaped call and assert the corresponding current-run verdict/block result (including a negative bypass case), so a stale or unrelated log cannot produce a false pass.

apurvvkumaria (changes requested):
Thank you for carrying the review feedback from #52 into a separate partner recipe. The x402 integration is in scope for this repository, but the security boundary and setup are not ready to merge yet. Please move the mandatory payment decision to OpenShell Supervisor middleware outside the agent sandbox, keep the decision genuinely pre-signature, provide a reproducible build/install/enable path, and make verification exercise a fresh payment call rather than accepting an existing log line. The current checks have not run because GitHub Actions is awaiting approval for this external contribution.

examples/recipes/partners/bluetier/x402-payment-gate/policy.yaml
```diff
@@ -0,0 +1,61 @@
...
+          path: /healthz
+      # x402 discovery document (lists the service's endpoints + pricing).
+      - allow:
+          method: GET
+          path: /.well-known/x402
+    binaries:
+    - path: /usr/bin/python3
+    - path: /usr/bin/python3.13
+    - path: /usr/local/bin/python3
+    - path: /usr/bin/curl
```
apurvvkumaria, line 146: The enforced endpoint allows only Python and curl, but the mandatory hook makes this request from the OpenClaw Node process. Node is therefore denied by this binary allowlist. In addition, `bring-up.sh` applies this file as the sandbox's complete policy even though it contains no inference route, so the advertised OpenClaw agent cannot operate normally. Please provide and verify a coherent full policy (or a supported composition step) that allows the exact runtime binary and retains the agent's required NemoClaw routes without broadening egress.

examples/recipes/partners/bluetier/x402-payment-gate/README.md
```diff
@@ -0,0 +1,219 @@
...
+flowchart LR
+    blackwall["External\nBlackwall verdict service\nblackwall-free.onrender.com"]
+    seller["External\nx402 seller\n(402 challenge)"]
+    chain["External\nx402 facilitator / Base\n(settlement -- NOT allowed\nby this policy)"]
+
+    subgraph host["Host Machine"]
+        subgraph sandbox["OpenShell Sandbox"]
+            agent["Hermes Agent"]
+            skill["blackwall-payment-gate\nskill + stdlib client\n(advisory)"]
+            hook["blackwall-x402-gate plugin\nbefore_tool_call hook\n(mandatory)"]
```
apurvvkumaria: The diagram and skill configure a Hermes agent, while the mandatory hook is an OpenClaw plugin. These are different NemoClaw agent runtimes, and the PR does not provide a supported deployment in which both layers run together. Please choose a coherent runtime or provide separate, independently runnable variants with the correct integration and lifecycle for each.

examples/recipes/partners/bluetier/x402-payment-gate/plugin/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "blackwall-x402-gate-example",
+  "private": true,
+  "version": "0.1.0",
+  "type": "module",
+  "scripts": {
+    "test": "vitest run --exclude '**/eval-live.test.ts'",
+    "eval:live": "vitest run eval-live.test.ts"
+  },
+  "devDependencies": {
+    "vitest": "^3.0.0"
```
apurvvkumaria: This adds Vitest and a committed dependency lockfile, but the PR description says no third-party dependency was added and `THIRD-PARTY-NOTICES` is unchanged. The contribution requirements require dependency and distributed third-party-content changes to be reflected there. Please update the notices and PR compliance statement, including any required open-source review disposition.

apurvvkumaria (commented):
Additional policy, runtime, and dependency blockers found during the same review:

examples/recipes/partners/bluetier/x402-payment-gate/host/release_gate.py
```diff
@@ -0,0 +1,326 @@
...
+            self._json(200, record)
+            return
+        self._json(404, {"error": "not found"})
+
+    def log_message(self, fmt, *args):  # quiet default access log
+        pass
+
+
+def main():
+    server = ThreadingHTTPServer(("127.0.0.1", GATE_PORT), Handler)
```
apurvvkumaria: The sandbox connects to `host.openshell.internal:8790`, but this server listens only on host loopback. For Docker/Podman-style drivers, the host alias resolves a bridge/non-loopback address; a service bound to `127.0.0.1` is not reachable there. OpenShell's own host-service example requires a suitable non-loopback bind for container reachability. Please make the gate bind address explicit and safely configurable, bind it to the interface the sandbox can actually reach, and prove the connection with the real in-sandbox request while retaining the narrow L7 policy.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/verify.sh
```diff
@@ -0,0 +1,97 @@
...
+
+echo
+echo "== 3/3 denied edge (in-sandbox; needs openshell + SANDBOX_NAME) =="
+if [ -z "${SANDBOX_NAME:-}" ] || ! command -v openshell >/dev/null 2>&1; then
+  echo "skipped: set SANDBOX_NAME and run where the openshell CLI is available."
+  echo "         Inside the sandbox, these two must hold:"
+  echo "           curl http://host.openshell.internal:8790/healthz   -> 200 (maker path)"
+  echo "           curl http://host.openshell.internal:8780/healthz   -> DENIED by policy"
+else
+  if openshell sandbox exec --name "$SANDBOX_NAME" -- \
```
apurvvkumaria: This still checks only `/healthz`; it does not submit the fresh in-sandbox intent promised by the README and commit description. It also skips the entire sandbox stage unless `SANDBOX_NAME` is manually exported, even though bring-up creates the default `x402-gate-demo`. Please use the same default here, POST a uniquely identifiable intent from inside that sandbox, and assert the current response/id before testing that the rail remains denied.

examples/recipes/partners/bluetier/x402-payment-gate/host/release_gate.py
```diff
@@ -0,0 +1,326 @@
...
+            code, err = approve_record(
+                record, self.headers.get("X-Operator"),
+                self.headers.get("X-Approve-Token"), APPROVE_TOKEN)
+            if code:
+                self._json(code, {"error": err})
+                return
+            # Named-human override of a HOLD: sign-then-settle now. STOPs can
+            # never reach here (status "refused" is terminal).
+            operator = self.headers.get("X-Operator")
+            signature = simulate_signature(record["intent"])
```
apurvvkumaria: The documented invariant says the gate settles only after a fresh GO, but this approval branch signs and settles an earlier HOLD without obtaining another verdict. The token authenticates an override; it does not turn HOLD into GO. Please choose and document one contract: either re-forecast and require a fresh GO before signing, or explicitly describe authenticated human approval as a separate override to the GO-only rule. Add handler-level tests for whichever contract is intended.

examples/recipes/partners/bluetier/x402-payment-gate/host/release_gate.py
```diff
@@ -0,0 +1,326 @@
...
+            return
+        m = re.match(r"^/v1/intents/([0-9a-f-]+)$", self.path)
+        if m and m.group(1) in INTENTS:
+            self._json(200, INTENTS[m.group(1)])
+            return
+        self._json(404, {"error": "not found"})
+
+    def do_POST(self):
+        if self.path == "/v1/intents":
+            if not store_has_room(INTENTS, MAX_INTENTS):
```
apurvvkumaria: The new capacity check is still raceable and therefore does not bound memory or verdict-service load as claimed. Under `ThreadingHTTPServer`, many concurrent requests can all observe room, then spend up to 90 seconds forecasting before any record is inserted at line 269. Reserve capacity atomically under the lock before starting the forecast (and release the reservation on failure), or use an equivalent bounded primitive, with a concurrency test proving submissions cannot exceed the cap.

apurvvkumaria (commented):
Thank you for the substantial hardening in the latest commit. The approval token, double-release guard, and input bounds are useful improvements. Four merge blockers remain in the current lifecycle and enforcement claims:

examples/recipes/partners/bluetier/x402-payment-gate/scripts/verify.sh
```diff
@@ -0,0 +1,116 @@
...
+# rail ledger. Stale lines cannot satisfy this.
+if [ -n "$TX1" ] && ledger_has_tx "$TX1"; then
+  echo "  PASS: this run's settlement $TX1 is on the rail ledger"
+else
+  echo "  FAIL: this run's settlement tx not found on the rail ledger"; FAIL=1
+fi
+
+echo
+echo "== 3/3 in-sandbox maker path + denied edge (needs openshell + SANDBOX_NAME) =="
+if [ -z "${SANDBOX_NAME:-}" ] || ! command -v openshell >/dev/null 2>&1; then
```
apurvvkumaria: The normal bring-up path creates `x402-gate-demo` but does not export `SANDBOX_NAME`, so running `verify.sh` as documented skips all three sandbox-boundary assertions. Please use the same default as `bring-up.sh` and `tear-down.sh` (`SANDBOX_NAME="${SANDBOX_NAME:-x402-gate-demo}"`) and skip only when OpenShell is unavailable. This prevents verification from reporting success without testing the maker route and denied rail edge.

apurvvkumaria:
DCO is still outstanding: none of the four commits currently in this PR includes a `Signed-off-by:` trailer. The latest commit has a verified cryptographic signature, but that does not replace DCO sign-off. Please amend or rebase all four commits with a valid sign-off from the contributor certifying each commit, while preserving the original authorship and co-author credit.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/bring-up.sh
```diff
@@ -0,0 +1,76 @@
...
+  echo "  the host boundary; re-run bring-up where openshell is available for"
+  echo "  the full sandbox."
+  exit 0
+fi
+if openshell sandbox list 2>/dev/null | grep -qE "^\s*$SANDBOX_NAME\s"; then
+  echo "  sandbox exists; re-applying policy"
+else
+  openshell sandbox create \
+    --name "$SANDBOX_NAME" \
+    --image "$IMAGE_TAG" \
```
apurvvkumaria: This lifecycle cannot create the advertised sandbox: OpenShell 0.0.85 rejects `--image`; `sandbox create` accepts `--from`. After correcting the source flag, bring-up must also configure or verify inference and start Hermes (`nemoclaw-start` or a documented equivalent). Without a command, OpenShell defaults to an interactive shell, so this does not produce the usable Hermes agent described by the README. Please make clean-checkout bring-up start and health-check the complete runtime.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/bring-up.sh
```diff
@@ -0,0 +1,76 @@
...
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+EXAMPLE_DIR="$(dirname "$DIR")"
+RUN="$EXAMPLE_DIR/.run"
+SANDBOX_NAME="${SANDBOX_NAME:-x402-gate-demo}"
+IMAGE_TAG="${IMAGE_TAG:-x402-payment-gate-sandbox}"
+mkdir -p "$RUN"
+
+echo "== 1/3 host services (the CHECKER side) =="
+command -v python3 >/dev/null || { echo "python3 required" >&2; exit 1; }
+if curl -sS -m 2 http://127.0.0.1:8780/healthz >/dev/null 2>&1; then
```
apurvvkumaria: Please do not accept any process that answers `/healthz` as the service from this checkout. A prior checkout or unrelated process on these ports is silently reused, so fresh verification can exercise stale code. Track and validate the processes that this lifecycle owns, restart its stale instances, or fail with an actionable port-conflict error.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/verify.sh
```diff
@@ -0,0 +1,137 @@
...
+  echo "verify: FAILURES above"
+elif [ "$SANDBOX_EDGE_TESTED" -eq 1 ]; then
+  echo "verify: OK (host boundary + in-sandbox maker/denied-edge exercised)"
+else
+  # Never let a skipped security-critical stage read as a full pass.
+  echo "verify: PARTIAL — host boundary OK, but the in-sandbox maker/denied-edge"
+  echo "        was NOT exercised (see stage 3). Re-run on the host with openshell"
+  echo "        for the full check."
+fi
+exit "$FAIL"
```
apurvvkumaria: When OpenShell is unavailable, the denied-edge stage is skipped but `FAIL` remains zero, so this final command reports success to CI. The denied rail edge is the central security property and the default full verification must exit nonzero when it is not exercised. If a host-only check is useful, make it an explicit mode with a clearly distinct result.

examples/recipes/partners/bluetier/x402-payment-gate/host/release_gate.py
```diff
@@ -0,0 +1,379 @@
...
+BLACKWALL_URL = os.environ.get(
+    "BLACKWALL_URL", "https://blackwall-free.onrender.com")
+RAIL_URL = os.environ.get("RAIL_URL", "http://127.0.0.1:8780")
+GATE_PORT = int(os.environ.get("RELEASE_GATE_PORT", "8790"))
+# The gate must be reachable from the sandbox (via the host.openshell.internal
+# route in policy.yaml), so it CANNOT bind host loopback. Default 0.0.0.0; set
+# RELEASE_GATE_BIND to the specific host-internal interface in production. The
+# RAIL stays 127.0.0.1 (host-only) -- that asymmetry IS the denied edge: the
+# gate accepts intents (least privilege), only the host reaches settlement.
+GATE_BIND = os.environ.get("RELEASE_GATE_BIND", "0.0.0.0")
```
apurvvkumaria: The default `0.0.0.0` listener exposes submission, status, and `/approve` endpoints on every host interface. The OpenShell policy restricts sandbox traffic; it does not protect these endpoints from other host-network clients. Bind the sandbox-facing operations to a specific reachable interface or split the listeners, keep approval host-only, and verify both intended reachability and unintended denial.

examples/recipes/partners/bluetier/x402-payment-gate/README.md
```diff
@@ -0,0 +1,200 @@
...
+## Key invariants
+
+- The agent can screen, explain, and submit intents; it cannot sign, release,
+  or settle a payment. No signing key exists inside the sandbox (in this demo
+  no real key exists anywhere — the rail signature is an explicitly labeled
+  simulation).
+- The sandbox has no route to the rail or any facilitator, even when a tool
+  or prompt attempts one. Its egress is the inference endpoints, the verdict
+  service, and the intent-submission route — nothing else.
+- Only the host-side gate settles, only after a fresh GO verdict; verdict →
```
apurvvkumaria: This `fresh GO` invariant does not match `process_approval`, which signs and settles after a fresh `HOLD` when a named operator supplies the approval token. Please choose one contract: require a fresh GO before every release, or define human-approved HOLD as a separate authorization path. Then make the README, diagram, module documentation, PR text, and tests state the same rule.

apurvvkumaria (changes requested):
Thank you for the substantial rewrite. DCO, authorship, pre-signature claim handling, runtime consistency, dependency removal, fresh-call verification logic, and concurrency controls are now addressed. Four merge-blocker categories remain in the current lifecycle, verification result, listener boundary, and release contract; I have identified them inline. At 4a0c38a, the 26 current unit tests, shell syntax, tracked-file license headers, label taxonomy, and diff checks pass. Non-blocking: please update the PR description after the code changes because it still reports 38 plugin tests, while the plugin is no longer in this PR and the current host suite runs 26 tests. GitHub workflows are still awaiting external-contributor approval.

bluetieroperations-create:
Thanks for the thorough round — all six are addressed, history collapsed to a single verified commit with the DCO sign-off in both the commit trailer and the PR description.

1. **PR description** — corrected: 27 stdlib host tests, no plugin tests (the OpenClaw plugin lives in the Blackwall repo, contributed separately in #52).
2. **`bring-up.sh`** — now `openshell sandbox create --from sandbox/Dockerfile` (not `--image`), waits for Ready, and probes `sandbox exec`. Scoped honestly: the security boundary needs the created sandbox + policy, and `verify.sh` exercises the denied edge via `sandbox exec` (curl/python3 inside the sandbox) — it doesn't depend on a running agent. I'm validating the full bring-up on a real OpenShell host and will follow up here.
3. **Process ownership** — `bring-up.sh` reuses only a service it started (pid-file tracked) and refuses a port held by an unknown process, so fresh verification can't exercise stale code.
4. **Bind** — reverted the `0.0.0.0` overcorrection. Two listeners now: a sandbox-facing submit/status listener (`RELEASE_GATE_BIND`, default `127.0.0.1`) and a host-only approve listener on `127.0.0.1:8791`. Exposing submit can no longer expose `/approve`.
5. **`verify.sh`** — exits non-zero when the in-sandbox denied-edge stage isn't exercised, so a skip can't read as a pass; a host-only run is an explicit `VERIFY_HOST_ONLY=1` opt-in with a distinct result.
6. **Release contract** — defined two explicit authorization paths, stated identically in the README, module docstring, and tests: automated release requires a fresh GO; human-approved release is a separate path (named human + host-only token) that re-screens and still hard-blocks a fresh STOP. STOP is terminal on both.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/bring-up.sh
```diff
@@ -0,0 +1,190 @@
...
+  echo "  skipped: docker not available (only used for the optional local"
+  echo "  sanity build; OpenShell builds from --from in phase 3)."
+fi
+
+echo
+echo "== 3/3 sandbox ($SANDBOX_NAME) =="
+# What this phase stands up is the security boundary: a sandbox with
+# policy.yaml applied. The denied-edge and in-sandbox maker checks in
+# verify.sh run via `openshell sandbox exec` (curl/python3 straight inside the
+# sandbox) and therefore do NOT need the Hermes agent runtime to be running --
```
apurvvkumaria, line 131: `bring-up.sh` does not start or health-check Hermes. It creates a sandbox and tests raw `curl` and Python calls instead. A user following this Hermes recipe still cannot reach the advertised maker workflow. Make bring-up start the supported Hermes runtime and verify a payment intent through it, or re-scope the example and documentation as a lower-level boundary demonstration.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/bring-up.sh
```diff
@@ -0,0 +1,190 @@
...
+# Relay+Hermes image and `nemoclaw-start`; that is an operator step layered on
+# top, not part of the boundary this recipe verifies.
+if ! command -v openshell >/dev/null 2>&1; then
+  echo "  skipped: openshell not in PATH. Run scripts/verify.sh now to exercise"
+  echo "  the host boundary; re-run bring-up where openshell is available for"
+  echo "  the full sandbox + denied-edge test."
+  exit 0
+fi
+if openshell sandbox list 2>/dev/null | grep -qE "^\s*$SANDBOX_NAME\s"; then
+  echo "  sandbox exists; re-applying policy"
```
apurvvkumaria: Reusing any sandbox with this name and only reapplying the policy can retain an older Dockerfile, skill, or client. The current verification then exercises `curl`, so it cannot detect stale recipe content. Recreate or rebuild the sandbox for this checkout, or verify its image and source identity before reuse.

examples/recipes/partners/bluetier/x402-payment-gate/scripts/verify.sh
```diff
@@ -0,0 +1,158 @@
...
+  #    boundary CANNOT be judged by curl's exit code (a denial and a success both
+  #    exit 0). It is OPEN only if the sandbox received the rail's ACTUAL reply
+  #    (the mock-rail role marker); a policy denial, a connection refusal, or any
+  #    other response all mean the boundary held.
+  RAIL_REPLY=$(openshell sandbox exec --name "$SANDBOX_NAME" -- \
+    curl -sS -m 10 http://host.openshell.internal:8780/healthz 2>/dev/null || true)
+  if printf '%s' "$RAIL_REPLY" | grep -q "mock-rail"; then
+    echo "  FAIL: sandbox reached the RAIL — the denied edge is open! ($RAIL_REPLY)"; FAIL=1
+  else
+    echo "  PASS: rail route denied from the sandbox — boundary held (reply: ${RAIL_REPLY:-<blocked>})"
```
apurvvkumaria: This passes for every response except the mock-rail marker, including a DNS failure, timeout, connection refusal, proxy error, or malformed output. Those failures do not prove that OpenShell enforced the policy. Capture the HTTP status and body, require the expected `403 policy_denied` response, and fail on transport errors or unexpected responses.

examples/recipes/partners/bluetier/x402-payment-gate/README.md
```diff
@@ -0,0 +1,217 @@
...
+- **Replace the simulation at the edges, keep the boundary.** The mock rail
+  and simulated signature are demo stand-ins; production adopters wire the
+  release gate to a real wallet/signing provider so the same verdict gates
+  real signing — see the wallet and OpenClaw adapters in the Blackwall
+  repository (`integrations/wallets/`, `integrations/openclaw/`). For
+  OpenClaw-runtime agents, the related in-sandbox defense-in-depth hook is
+  contributed separately in the
+  [BLACK_WALL Preflight Guardrail](https://github.com/NVIDIA/nemoclaw-community/pull/52)
+  example.
+- **Supervisor middleware is a compatible tightening.** OpenShell's
```
apurvvkumaria: The earlier change request required OpenShell Supervisor middleware for the mandatory decision. This patch explicitly defers that work and substitutes a host-side release gate. The gate is stronger than an in-sandbox hook, but it does not resolve that architecture request. Implement the middleware, or obtain explicit maintainer agreement that this host-side boundary is accepted for this recipe and update the scope accordingly.

apurvvkumaria (changes requested):
Thank you for the substantial update. The signed-off commit, pre-signature host gate, split approval listener, concurrency controls, and fresh canaries address most of the prior feedback. All 27 unit tests and the repository checks pass locally. Four blockers remain in the current patch; see the inline comments. GitHub also reports no checks for the latest commit, so the workflows must run before merge.

bluetieroperations-create:
Thanks — all four addressed, validated end to end on a live OpenShell + Docker host, single verified commit with the DCO sign-off.

1. **Hermes runtime (re-scope).** Took the re-scope path: the README now states plainly that this is a host-side payment *boundary* demonstration. bring-up stands up the boundary and verify exercises the in-sandbox maker path at the intent-submission level (POST /v1/intents over the scoped route → verdict) plus the denied edge — it does not start an interactive Hermes agent, and the skill is documented as "how a Hermes agent would integrate," with the full Relay+Hermes runtime called out as a separate operator step.

2. **Sandbox reuse / stale content.** bring-up now deletes any existing sandbox of this name and rebuilds it from the current checkout on every run, so it can never exercise a stale Dockerfile/skill/client.

3. **Denied-edge assertion.** verify.sh now captures both the HTTP status and body and positively requires `403 policy_denied`. A rail breach, a DNS failure, a timeout, a connection refusal, a proxy error, or any unexpected/empty response all fail — none are treated as proof of enforcement. (Confirmed on a real host: the sandbox→rail request returns HTTP 403 policy_denied.)

4. **Supervisor middleware vs host-side gate.** Fair point that this shouldn't be a silent substitution. The README now frames it explicitly: the mandatory decision runs in a host-side release gate — the established pattern of the payment-ops-hermes recipe — with Supervisor middleware documented as the ideal fail-closed tightening and the next step, once that configuration surface is available to community recipes. Rather than defer unilaterally, I'd like to ask directly: **would the maintainers accept the host-side gate as the enforcement architecture for this recipe** (matching payment-ops-hermes), with the middleware tracked as follow-up? Happy to align on that before it's considered settled.

While validating on a real OpenShell host I also caught and fixed several integration issues the unit tests couldn't surface (Dockerfile build context under `--from`, the submit listener's bridge bind, a create-exit under `set -e`, and pid-file capture in tear-down), so the full bring-up → verify → tear-down lifecycle now runs clean.

apurvvkumaria (approved):
Approved. We accept the host-side release gate as the enforcement architecture for this boundary-focused example, with Supervisor middleware documented as a follow-up. The contributor addressed the lifecycle, stale-sandbox, denial-verification, DCO, and prior review issues. All required checks pass.

Issue #122, opened 2026-08-18
[Feature]: Add a memory-driven chief-of-staff example backed by a maintained personal memory


rebelle5868:
## Problem

A personal assistant is only as good as what it remembers about one person: who they work with, what they are accountable for, what they already decided, and what they consistently ignore.

Agent runtimes give you a place to put that, but it is a notepad. Hermes keeps memory in `memories/MEMORY.md` and `USER.md`, entry-delimited stores of short declarative notes the agent writes for itself. Nothing validates an entry, indexes it, links it to related entries, ages it, or repairs it. Over weeks it either bloats or drifts, and nothing detects either, because there is no contract to check the store against.

A second kind of thing also needs remembering, and it is not a fact. It is a judgment about an item that some other system owns: this message needs a reply, it ranks third this week, it is snoozed until Thursday, and it was demoted because the user overrode the ranking twice. A mailbox has no field for that. A Slack workspace has no field for that. Writing it back into either one mutates the user's real data to store the agent's opinion.

We checked the catalog before proposing. Each existing piece holds part of this and none holds the whole:

| Existing example or primitive | What it already does | What it does not cover |
| --- | --- | --- |
| Hermes `memories/` (platform primitive) | A durable place the agent can append notes to, preserved across profile updates | Flat and entry-delimited: no schema, no index, no cross-references, no decay, no provenance requirement, and nothing repairs it |
| `recipes/nvidia/developer-community-chief-of-staff` | Synthesizes signals into briefs and priorities, and records a per-item disposition | Its durable data is a re-scrapable community mirror, and the per-item disposition is a `PATCH isRead` into the source mailbox, so it is limited to the fields the source defines |
| `recipes/partners/tavily/watchtower` | A durable per-item ledger, advanced only after outputs are durable | Monotone and terminal: an item once seen is dropped forever, so nothing is re-judged |
| `recipes/community/deep-research-worker` | SQLite with a mutable per-item status machine, atomic claiming, and retries | Execution states over tasks the agent submitted itself, not judgments about items an external system owns |
| `recipes/nvidia/pr-review-advisor` | Human corrections persisted with provenance and recalled in later runs | Scoped to one review artifact per run, stored in the runtime's own memory, with no schema, no population ranking, and no repair schedule |

## Proposal

Add a chief-of-staff example whose distinguishing asset is a maintained memory of one person, and whose output quality follows from it. Every surface reads from the same store instead of re-deriving from raw sources.

[...]

PR #123, opened 2026-08-19
feat(examples): add memory-driven chief-of-staff store and tests

rebelle5868:
## Related Issue

Part of #122. This is the first of the three sequential phases agreed there; it does not close the issue.

## Description

Adds `examples/recipes/nvidia/memory-driven-chief-of-staff/`, the fixture-backed store and its tests.

The recipe keeps a locally-authoritative, revisable record per inbound email and Slack message, re-judged on a schedule and re-ranked under fixed caps, without ever writing back to the source. This phase contains the store, the memory schema, the deterministic checks, the five skills, a synthetic fixture corpus, a walkthrough that runs the whole mechanism offline, and the acceptance criteria from the issue. It reaches no network, requires no credentials, and integrates with no scheduler.

Two design decisions are worth flagging for review, because both moved work out of prompts and into code:

- **Tier assignment is arithmetic, so it lives in a pure function.** Hard caps of ten high and ten medium over the whole open population, an intent gate on the top tier, and cascade for un-gated rows that ranked into it. The visible consequence in the fixtures is a message that shouts — mandatory deadline, URGENT in the subject — capped at the middle tier because it matches nothing the user chose to work on.
- **Detection is mechanical even when the fix is not.** Four acceptance criteria could not be tested while the logic lived only in skill prose, so the checkable half was extracted: `memory_check` reports invariant violations, `preferences` counts corrections against a fixed threshold, `migrate` handles schema versioning. The skills keep the judgment.

**Naming.** `memory-driven-chief-of-staff`, per your decision on #122. The alternative we had raised is withdrawn.

## Review findings addressed

Both review rounds are addressed on this head. Each fix was verified by reproducing the failure first.

| Finding | Where |
| --- | --- |
| Caps and ranks computed per envelope, not across the open population | `ranking.rank_population`, unique index in `schema.sql`, migration to v2 |
| Schema-version check never reached from the real store-opening path | `_db.ensure_store` now migrates on open; tested through the writer's CLI |
| No required-frontmatter validation | `memory_check.check_frontmatter`, with a negative test per required field |
| Fixture corpus stopped at ingestion | `walkthrough.py`, plus `correct.py` — the user-correction writer that did not previously exist |
| `obligation-review` did not read the preference policy | `skills/obligation-review/SKILL.md` |
| Profile-home guard rejected the default profile | `_db._is_profile_home`, covering the default and named layouts |
| `DIST_OWNED` hardcoded while claiming to read the manifest | `test_durability.read_manifest`, plus an installation smoke test |
| A non-existent `HERMES_HOME` bypassed the profile check and was materialised | `_db.ledger_path` refuses a missing home |
| A directory holding only a `.DS_Store` was refused as occupied | `_db._significant` ignores dotfiles |
| The future-schema refusal ran after the baseline DDL | `migrate.refuse_if_from_the_future`, called before any DDL |
| A v1-to-v2 upgrade written against a v1 that never shipped | Ships at schema v1; the leaked `deleted_at` column is removed |
| Pinned rows bypassed the caps entirely | `ranking.rank_population` counts pins against the caps, overflow cascades |
| Repeated corrections counted as independent evidence | `correct.py` records only state transitions |
| Rerank audit covered only the envelope's own rows | `_rerank_all` takes prior state from the store |

Three defects surfaced while doing the above, all fixed here: a user pin ordered a row without deciding its tier and spent a capped slot on the way down; a single repeatedly-ignored sender was promoted into a rule covering their whole mail domain, which would have suppressed the user's own colleagues; and each test file's `__main__` guard sat mid-file, so classes appended below it were skipped when the file was run directly.

Two README claims are narrowed to match the decisions on #122: source-deletion reconciliation is scoped to Microsoft Graph with the Slack limitation documented, and owner-only permissions are stated as a filesystem access control rather than encryption at rest, with an encrypted volume or an application-level design named as a prerequisite before a connector stores real message bodies.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — all checked source files have SPDX headers.
- [ ] `python3 scripts/check_label_taxonomy.py --check` when governance metadata changes — not applicable; this branch changes no governance metadata.
- [x] Relevant example setup or syntax checks:
  - `fail=0; for t in tests/*.py; do python3 "$t" || fail=1; done; echo "failed=$fail"` from `profile/scripts` — 157 tests, every file `OK`, `failed=0`. Discovery reports the same 157, and a test asserts the two counts cannot diverge. The earlier `|| break` form was wrong: a `for` loop reports its last command's status, so a failing test still left the loop exiting `0`.
  - `HERMES_HOME=$(mktemp -d)` then `python3 profile/scripts/walkthrough.py --fixtures fixtures` from the recipe root — runs ingestion, judgment, two corrections, a re-judgment, the preference threshold and the memory self-check, including one deliberate break so the check is seen to fail as well as pass.
  - `python3 profile/scripts/load_fixtures.py --fixtures fixtures` twice — first run stores 8 records, second stores 0, so intake is idempotent on the source's own id.
  - `python3 profile/scripts/memory_check.py` — reports the seeded memory clean against its own schema.

**Runtime validation.** The profile was installed into a Hermes 0.19.0 runtime and the intake job run against a real model. That is what found the three defects reported in the previous comment, including the one that invalidated the central claim while every test stayed green. The profile-home fix on this head was checked against a real `~/.hermes` layout, which the previous guard refused. Not re-run end to end on this head: nothing here has a network or credential path, and the runtime's own preservation behaviour is stated in the README as measured separately rather than reproducible from this branch.

- Result: `docs-updated`
- Reviewer: Apurv Kumaria (@apurvvkumaria), final maintainer review
- Evidence or justification: `README.md` and `fixtures/README.md` were reviewed against `WRITING.md`, the controlled-word list, and CONTRIBUTING's twelve items for a new example, over eight rounds. Each round ran every command both documents give, from the directory each names. The rounds were not cosmetic: they found and corrected the walkthrough recording two model turns while both documents said one; a verification command that exited `0` when a test failed; a claim that the seed memory drives the intent gate in the fixture path when it does not; two claims about Hermes contradicted by Hermes's own source and documentation; two shipped skills reading a path nothing creates; a `workspace` update hazard stated backwards; and a source-write scan described as a guarantee when it matches call shapes. A separate reviewer working from the shell independently reached two of the same findings and additionally exercised `hermes profile install --force` and `hermes profile update` against a real runtime, which is the one claim in this README no earlier round could check. On this head I re-ran the documented verification myself: the walkthrough prints seven steps and exits `0`; `load_fixtures.py` reports `"added": 8` then `"added": 0`; `memory_check.py` exits `0`; the documented test loop reports 157 tests with `failed=0`, matching discovery; every documented exit code reproduces (`3` on an ignored row and an unknown id, `0` on a repeat, `1` with `HERMES_HOME` unset, `2` from `memory_check.py`, `2` on a second walkthrough run); all four external links return `200`; `check_license_headers.py --check` passes and `git diff --check` is clean. Two follow-ups are recorded rather than fixed here, both narrow: the recipe's `.py` comments still carry a few British spellings, and `preferences.collect` joins three tables, which suits this scale but is noted in its docstring as unsuitable for a store whose tables are id-linked silos.

## Release And Compliance

- [x] Public content uses sanitized examples and placeholders instead of private values. The fixture corpus is written from scratch: the people, the company, the projects, and every message body are invented rather than anonymized, and `fixtures/README.md` carries an in-band provenance block saying so.


apurvvkumaria (changes requested):
Thank you for the audit update. The repository checks and all 59 recipe tests pass, but the following implementation and documentation gaps still need to be addressed before approval:

1. **Enforce priority caps and ranks across the full open population.** `apply_decisions.py` assigns tiers only across the rows in the current envelope. Applying two separate 20-row envelopes leaves 20 `high` and 20 `medium` obligations, with duplicate `global_rank` values. The stated 10/10 caps therefore do not hold across intake or review batches. Recompute or validate the complete active ranking transactionally, and add a regression test that applies multiple batches and asserts the global caps and rank uniqueness.

2. **Wire schema-version protection into the real store-opening path.** The executables call `ensure_store()`, which executes `schema.sql` but never calls `migrate()` or `open_store()`. A database marked with schema version 99 is accepted and populated with the current tables instead of being refused. Use one initialization/opening path that creates a new store, migrates supported older stores, and rejects future versions before any write. Test this through the writer or CLI path, not only by calling the migration helper directly.

3. **Make the deterministic checker enforce the documented schema contract.** `memory-repair/SKILL.md` says frontmatter completeness is a mechanical finding, but `memory_check.py` has no frontmatter-completeness check. Removing the required `name` field from a person page still produces `clean: true`. Add required-field validation for the index and each page type, with negative tests, or narrow the documentation so it does not claim this is enforced.

4. **Make the fixture walkthrough exercise the behavior it promises.** The documented loader creates eight `items` but zero `obligations` and zero audit `events`; it does not demonstrate ranking, the intent gate, re-ranking, or absorbing a correction. Add a deterministic decision/correction fixture and runnable commands that expose those results, or revise the README and fixture documentation to describe only ingestion and normalization.

5. **Align preference-policy consumption.** `preference-update/SKILL.md` says both later flows read `workspace/policy/preferences.md`, but `obligation-review/SKILL.md` does not instruct the agent to read it. Add the policy to the review skill's required inputs, or correct the stated behavior.

After these changes, please complete the documentation-writer review on the final head and replace the controlled-word-list uses of “demo” in `fixtures/README.md` with “example.”

rebelle5868:
Pushed three fixes found by installing this profile into a Hermes 0.19.0 runtime and running the intake job against a real model. None of them was reachable from the test suite, and one of them invalidated the central claim while every test stayed green — so it seems worth writing down rather than just fixing.

**The profile had no `distribution.yaml`.** It could not be installed at all. Sixty-one tests passed against something that would not start, because they exercise the scripts directly and never install anything. Added, along with the `SOUL.md` the persona was missing. `cron` is deliberately absent from `distribution_owned`, so an update cannot replace a live job store along with its run history.

**Every memory path in the skills was relative, and the agent's working directory is not the profile home.** All ten resolved to nothing. The failure mode is the one worth noting: an unreadable memory is indistinguishable from an empty one, so the intent gate did not error — it silently stopped gating, and every row landed in the middle tier. The run looked successful and the ranking looked merely conservative.

With absolute paths, the same eight fixtures produce the behaviour this recipe exists to demonstrate:

```
#1  high    gated    "Billing migration — need the cutover window confirmed"
#4  medium  un-gated "URGENT: expense policy attestation closes Friday"
```

The second is mandatory, has a hard deadline, and says URGENT in the subject. It is capped below a quietly-worded request, because the first maps to an entry in the user's own priorities page and the second matches nothing they chose. That is the claim, now observed rather than asserted.

**A destructive command defaulted to the runtime root.** With `HERMES_HOME` unset, `ledger_path()` fell back to the default profile, and a reset reported deleting a ledger belonging to a different profile entirely. Local tests always set the variable, so the default was never exercised. It now refuses a path that is not a profile home.

Also: the checks skip AppleDouble sidecars. They are binary, they end in `.md`, and they killed the whole pass when a macOS archive carried them along — which will happen to anyone who packages this from a Mac.

The two credential-custody questions above are still open and still block phase 3 only.

apurvvkumaria (changes requested):
Thank you for testing the profile in a real Hermes runtime. The latest commit improves installability, but two additional issues need to be addressed:

1. **Do not reject the valid default Hermes profile.** `ledger_path()` currently rejects every `HERMES_HOME` whose final path component is `.hermes`. Hermes supports `~/.hermes` as the default profile home: https://github.com/NousResearch/hermes-agent/blob/main/website/docs/user-guide/profiles.md. Requiring an explicit `HERMES_HOME` is reasonable, but the basename does not distinguish a runtime root from a valid profile. Please use a supported profile signal instead and add coverage for both the default and named-profile layouts.

2. **Make the distribution test validate the shipped manifest.** `test_durability.py` says it reads the real distribution manifest, but `DIST_OWNED` is hardcoded and already differs from `distribution.yaml` by omitting `distribution.yaml` itself. Please derive the assertion from the shipped manifest and add a deterministic installation smoke test that verifies the required profile files. This is important because the missing manifest and persona file were not detected while all existing tests passed.

The five items in the previous review also remain outstanding: global population ranking and caps, schema-version enforcement through the real store-opening path, required frontmatter validation, a deterministic ranking/correction fixture walkthrough, and preference-policy consumption during obligation review.

Please also refresh the PR description and documentation checklist on the final head, including the current test count and runtime-validation status, and resolve the remaining controlled-word-list findings.

rebelle5868:
Thank you — both rounds are addressed on `21c7540`, and the second round caught a defect I had introduced in the first. Each fix was verified by reproducing the failure before keeping the test.

## From the second review

**1. The guard rejected the default profile.** You were right, and the mistake was worse than the symptom: I had checked the last path component, and `~/.hermes` is the default profile home itself (`hermes_cli/profiles.py`: *"The 'default' profile is `~/.hermes` itself — backward compatible"*). The runtime root and a valid profile home can be the same directory, so no name-based test can separate them, and every default installation would have been refused.

Replaced with a marker check — `distribution.yaml` (the file Hermes itself reads to identify an installed distribution), `SOUL.md`, `.env`, `memories`, `sessions`, `skills`, `cron`. Any one is enough, since a profile part-way through setup is still a profile, and a fresh empty directory is accepted so a first run can create the store. Covered for the default layout, the named `<root>/profiles/<name>` layout, a home that merely ends in `.hermes`, each marker on its own, the fresh-directory case, an unset variable, a file, and a directory that is not a profile. I confirmed the new tests fail against the old check before keeping them, and resolved the real `~/.hermes` on this machine, which the previous version refused.

**2. The distribution test now reads the shipped manifest.** The old comment claimed it did while the next line restated it, and it had already drifted — omitting `distribution.yaml` itself, exactly as you said. It now parses the manifest (a small stdlib parser; the recipe depends on nothing outside it) and derives the assertion.

Added an installation smoke test over the artifact rather than the code: every key Hermes reads is present, every declared owned path is actually shipped, nothing at the profile root is undeclared, every skill has a `SKILL.md` with frontmatter, and a copy-and-open exercise of the installed layout. I verified it fails in both of the ways that previously passed silently — removing `distribution.yaml` from the manifest, and deleting `SOUL.md`.

**3. PR description and checklist refreshed** on this head, including the current count (110 tests) and an explicit runtime-validation note.

**4. Controlled words.** Re-checked the whole recipe against the list. The one real finding — `demo` where `example` is required — is fixed. The remaining matches are `demote` and `demonstrate`, a local variable named `sample` in test code, and `owner` used for a filesystem owner and for a project role in the memory schema; the list does not cover those. If you are seeing findings I am not, please name them and I will fix them.

## From the first review

**Global caps and rank uniqueness.** They were computed per envelope, so two batches each kept their own top ten and both claimed positions 1..20. `rank_population` now orders the whole open list, and a partial unique index on `global_rank` backs it in the schema (v2, with a migration that clears pre-existing duplicates). The regression test applies two twenty-row batches, gating six in the first and eight in the second — fourteen gate-passing rows, more than the tier holds, but neither batch exceeds the cap alone, so a batch-local implementation admits all fourteen. Measured against the old behaviour: batch-local gives `high=14` with duplicate ranks; global gives `high=10`, unique.

**Future-schema rejection on the real path.** `ensure_store` only ran `executescript` and never called `migrate`, so the guard lived in a helper no executable invoked. It now migrates on open and raises before any write, tested through `apply_decisions.py` over its command line, asserting a non-zero exit and that no row was written. The guard is one-sided: an older store is still migrated.

**Required frontmatter.** Added per-page-type required fields and enumerated values, both taken from `schema.md`. Removing `name` from a person page now reports `missing-field`. There is a negative test for every required field, dropped one at a time so no check can pass by way of another's finding, plus an out-of-range enum and a page with no frontmatter block. It runs inside `check_all`, so the scheduled repair job sees it too.

**The fixture walkthrough.** Fixed rather than narrowed, but the reason it was empty was deeper than thin fixtures: **there was no code path by which a correction could be recorded at all.** `preferences.py` counts only `actor='user'` events, and the skills correctly forbid the agent from writing them — so nothing could write one, and the policy loop had no way to start. This adds `correct.py`, the user's writer, which is the only producer of those events.

`walkthrough.py` then runs the mechanism with no credentials and no model. One recorded envelope stands in for a single model turn; everything downstream of it — gate, caps, ranking, writer, correction, re-ranking — is the shipped code, and the script says which is which on screen. Seven steps, ending with the memory self-check breaking a page on purpose so a reader sees the check fail as well as pass. `test_walkthrough.py` asserts each printed claim against the store it produced. It stops honestly short of writing a policy rule: two corrections do not reach the threshold of three, and rather than manufacture a third it prints the tally and says so.

**`preferences.md` in `obligation-review`.** Added, with the reason in the skill: that pass re-ranks, so omitting it is precisely how a correction gets undone.

## Defects found along the way

- `manual_priority` ordered a row but did not decide its tier, so a pin sent a row to the bottom while it kept the tier the gate gave it — and spent a capped slot on the way down.
- `candidates()` promoted a single repeatedly-ignored sender into a rule about their whole mail domain. Colleagues share the user's domain, so three ignores of one newsletter would have written a rule suppressing the highest-ranked people in the memory. A domain rule now needs a second sender.
- Each test file's `__main__` guard sat mid-file, so classes appended below it were never defined when `unittest.main()` ran. Twenty-one tests — every one added for the first review — were invisible to the way the README documents running them. There is now a test asserting no class sits below the guard, because I made this mistake twice.

## From your direction on #122

Two decisions landed in this branch's text rather than waiting for a later phase: source-deletion reconciliation is now scoped to Microsoft Graph with the Slack limitation documented, and owner-only permissions are stated as a filesystem access control rather than encryption at rest, with an encrypted volume or an application-level design named as a prerequisite before a connector stores real bodies.

The remaining decisions are phase-2 and phase-3 scope. I will update the issue body with all seven, and take the OpenShell OAuth2 refresh question for Slack as the next piece of work — confirming whether the existing strategy can handle single-use refresh-token rotation before proposing anything, and raising it upstream rather than working around it if it cannot.

apurvvkumaria (changes requested):
Thank you. The latest head resolves the default-profile guard, manifest drift, installation coverage, controlled-word finding, and stale PR description. All 110 recipe tests and the repository checks pass. I reproduced the following remaining issues against `21c7540`:

1. **Fix migration from the actual v1 schema and reject future schemas before any DDL.** Opening a database created from the prior v1 `schema.sql` fails with `OperationalError: no such column: deleted_at` before the v2 migration runs. `ensure_store()` executes the current schema first, whose `idx_items_deleted` index references a column the real v1 table does not have. The migration test instead makes a current database look old, so it retains `deleted_at` and misses this failure. The same ordering also creates the current tables in a schema-99 database before raising `SchemaFromTheFuture`. Please inspect the existing version before applying current DDL, migrate every column introduced after v1, and test from an actual v1 schema artifact. A future-schema test should assert that the database remains byte-for-byte or schema-equivalent unchanged, not only that no obligation row was added.

2. **Keep the documented hard caps when manual priorities are present, or explicitly narrow the contract.** Passing eleven rows with `manual_priority='high'` to `rank_population()` returns eleven high rows. The README and ranking module state a hard maximum of ten high and ten medium across the whole open population. Either include pinned rows in those caps with documented overflow behavior, or clearly state and test that user overrides can exceed them.

3. **Make correction commands idempotent.** Calling `correct.ignore()` three times for the same already-ignored obligation writes three user correction events. `preferences.collect()` then treats the retries as three independent corrections and produces a learned sender rule at the threshold. Only a real state transition should add evidence; repeated ignores should be a no-op or return a clear error. Apply the same state-transition validation to `unignore` and priority overrides where applicable.

4. **Audit every row changed by population reranking.** When a later batch displaces an untouched row from high rank 10 to medium rank 11, its tier and rank change but no `reranked` event is written. `_rerank_all()` captures prior state only for rows present in the current envelope. Snapshot every open row before recomputing and log each changed row so the stated audit trail covers population-wide effects.

The documentation-writer review also remains incomplete on the final head. Once these implementation issues are resolved, that review can be completed against the final text.

rebelle5868:
Follow-up on `6f27da8`. Reviewing the guards I added in the previous commit turned up three gaps in them, all reachable, so they are fixed here rather than left for a third round. Each has a test that fails without its fix.

**A `HERMES_HOME` that did not exist bypassed the profile check entirely.** Both branches were gated on the path already existing, so a mistyped path fell through and `ensure_store` went on to materialise a whole store under the misspelled name. The rule was inverted: a wrong directory that exists was caught, a wrong path that does not was obeyed — and a typo is the ordinary way of naming the wrong place, which is the case this guard exists for. A missing home is now refused, naming both supported layouts.

**A directory holding only a `.DS_Store` was refused as occupied.** The freshness branch treated any entry as content, so opening a folder in a file browser changed what the code thought it was. Dotfiles no longer count. The markers that are themselves dotfiles are matched earlier, and a directory with real unrelated content is still refused — there is a test for that, so the relaxation cannot quietly become a hole.

**The future-schema refusal ran after the baseline DDL.** Your original wording was "rejects future versions before any write", and it did not. `CREATE TABLE IF NOT EXISTS` is idempotent against our own schema but not a later version's, so a v99 store that had dropped a table this version still ships had it silently recreated on the way to being refused — leaving the store carrying a table from a schema the refusing code does not understand. The structural cause is that `current_version` reads `meta`, which a new store does not have, so the DDL had to come first. Split out into `refuse_if_from_the_future`, which probes for `meta` and returns quietly when it is absent, and is called before any DDL runs.

The last one is the one worth flagging: the previous test asserted that no rows were written, and DDL is not rows, so it stayed green while the store was being modified. The test now compares the full table list before and after.

117 tests, direct execution and discovery agreeing.

I have also updated #122 with your seven decisions, including the two that landed in this branch's text (Graph-scoped deletion reconciliation, and permissions stated as access control rather than encryption at rest). Next from me is the OpenShell OAuth2 determination for Slack under decision 3.

apurvvkumaria (changes requested):
Thank you. I confirmed the latest head fixes the future-schema ordering issue: a schema-99 database is now refused before baseline DDL and remains unchanged. The missing-home and dotfile profile checks also behave as documented. All 117 recipe tests and the repository checks pass.

The following items remain reproducible on `6f27da8`:

1. **Resolve the real v1 upgrade path.** Opening a database created from the prior v1 `schema.sql` still fails with `OperationalError: no such column: deleted_at` before the v2 migration runs. The migration test starts from the current schema and removes only `batch_rank`, so it retains `deleted_at` and does not represent v1. If no v1 schema has shipped, the clean solution may be to ship this new recipe as schema v1 and remove the unneeded upgrade claim. If v1-to-v2 migration is part of the contract, add a true v1 fixture and migrate every post-v1 column before creating indexes that reference them.

2. **Reconcile manual priorities with the documented hard caps.** Eleven rows pinned to `manual_priority='high'` still produce eleven high rows. Either keep the ten-item cap across pinned and unpinned rows with documented overflow behavior, or explicitly document and test that direct user overrides may exceed the cap.

3. **Make correction commands idempotent.** Three repeated `ignore` calls for the same already-ignored row still write three user correction events, and those retries reach the preference threshold. Only a real state transition should add evidence. Please cover repeated ignore, unignore, and unchanged priority overrides.

4. **Audit population-wide reranking completely.** A later batch can move an untouched row from high rank 10 to medium rank 11 without writing a `reranked` event. Capture the prior state of the whole open population and log every changed row, not only rows included in the current envelope.

The documentation-writer review also remains incomplete on the final head.

apurvvkumaria (changes requested):
Thank you. The latest commit resolves the four previous implementation findings. I confirmed that the current schema ships cleanly at v1, future schemas remain untouched, pinned rows stay within the caps, repeated corrections no longer reach the preference threshold, and agent-driven population reranking audits displaced rows. All 134 recipe tests and repository checks pass.

Two final correction-path issues remain reproducible on `1a31cb7`:

1. **Audit rows displaced by user corrections.** `correct._rerank_open()` changes the priority and global rank of the remaining population after an ignore, restore, or priority override, but it writes no `reranked` events for those affected rows. For example, ignoring rank 1 moves ranks 2 and 3 to positions 1 and 2 with no audit record. Apply the same before/after population audit used by `apply_decisions._rerank_all()` to the correction path, while avoiding duplicate events for the directly corrected row.

2. **Enforce valid correction state transitions.** `unignore()` currently reopens any row whose status is not already `open`, including a completed obligation: a `done` row becomes `open`. Restrict restore to `ignored -> open`; an already-open row can remain an idempotent no-op, and invalid states such as `done` should be rejected or left unchanged. Please apply the same explicit state rules to `ignore` and priority overrides so closed obligations cannot be silently rewritten as user-preference evidence.

The documentation checklist also needs correction. It marks the documentation-writer review complete while the result says `blocked` and no reviewer is named. Please either complete the review on the final head and identify the reviewer, or leave the completion box unchecked.

apurvvkumaria:
Thank you. The documentation update on `6cc70e6` addresses the previous overclaims: it discloses both recorded model turns, distinguishes the recorded gate verdict from live memory-driven inference, adds the no-gate comparison, and fixes the verification command so test failures propagate. I reran the documented suite: all 134 tests pass. DCO, SPDX, and governance checks also pass.

The following items still need correction before approval:

1. **Audit correction-driven population reranking.** `correct._rerank_open()` updates every open row's priority and global rank but does not snapshot the prior population or write `reranked` events. I reproduced this by ignoring rank 1: the remaining rows moved from ranks 2 and 3 to 1 and 2, while the number of `reranked` events remained zero. Please audit every indirectly changed row, without duplicating the directly corrected row's correction event, and add regression coverage for ignore, restore, and priority override paths.

2. **Enforce correction state transitions.** `unignore()` still reopens every non-open row, including completed obligations. I reproduced a `done` row changing back to `open`. `ignore()` and `set_priority()` also accept closed rows and can record invalid preference evidence. Please restrict restore to `ignored -> open`, treat valid repeated operations as no-ops, and reject or leave unchanged invalid transitions involving completed rows. Add tests for all three correction commands.

3. **Correct the documentation-review checklist.** The PR currently marks the review complete while reporting `Result: blocked` and leaving `Reviewer:` empty. Please either record the completed final-head review and reviewer, or leave the completion box unchecked.

The existing 134 tests do not exercise the first two cases above.

rebelle5868:
All items from both reviews are addressed. Current head is `45b51ff`; the branch also carries a merge of `main`, done as a merge rather than a rebase per the new "Update a Published Branch" section.

**Correction-path audit.** `_rerank_open` recorded nothing about the rows a correction displaced — the same defect fixed in the agent's writer last round, still present in the user's path. Displaced rows are now audited as `agent` rather than `user`: the user chose one row, not the reshuffle underneath it, and `preferences.collect` counts only user-authored rows. The corrected row keeps its own correction event and is not audited twice.

**Correction state transitions.** All three commands now refuse a `done` obligation and exit `3`; restore applies only to `ignored → open`; a pin applies only to the open list; a valid repeat stays a no-op. Writing the tests for that surfaced a third defect: `unignore` set `status='open'` while the row still carried the rank it held when ignored, so reopening tripped the uniqueness index against whichever row holds that position now — restore raised `IntegrityError` rather than working. The position is cleared on the way back in.

**Documentation.** I put the two READMEs through independent review, three rounds, each one running every documented command. It found more than I expected, including three claims that were simply false: the walkthrough records two model turns and both documents said one; the seed memory does not drive the gate in the fixture path, so deleting it leaves the printed tiers unchanged, while `fixtures/README.md` claimed the opposite; and the verification command I had documented (`|| break`) exits `0` when a test fails, so a reader following it would report a false pass. All three are fixed, and step 2 of the walkthrough now re-ranks the same rows with the gate verdicts withheld and prints `high=0`, so what the reservation buys is visible rather than asserted.

**On the documentation-writer review — I was wrong earlier.** I said the role was defined nowhere in the repository. I had searched the repository text and not the merged pull requests, where the convention is plain. That claim is withdrawn.

**One more, found while probing the correction path.** `write_txn` rolled back unconditionally, so when `BEGIN IMMEDIATE` failed on a busy database the rollback failed too and raised "cannot rollback - no transaction is active", discarding "database is locked". Every writer went through it. Fixed, with a contention test.

152 tests, direct execution and discovery agreeing. The audit-row growth that per-row auditing implies is now recorded under Known limitations: about 100 audit rows for one ignore on a 200-row open list, with `events` append-only and no compaction path in this phase.

apurvvkumaria:
Thank you. I reviewed `e964f8a` and reran the documented verification. The latest documentation changes correctly narrow the Hermes memory claim, document the exact `unignore` recovery command, separate implemented recipient reduction from future connector behavior, and scope the application-state write boundary accurately.

The substantive review findings are resolved. Both documented test methods report 152 passing tests, `git diff --check` is clean, all 19 checked source files have SPDX headers, and the required GitHub checks pass. The branch is current with `main` and mergeable. I found no new code or documentation blocker.

Please make the following final PR-description updates before approval:

1. Change the verification result from 134 tests to 152 tests.
2. Complete the documentation-writer record for this final head. The contributor comments describe multiple independent documentation reviews, but the PR description still says `Result: blocked`, leaves `Reviewer:` empty, and leaves the completion box unchecked. Record `docs-updated`, the reviewer, and concise final-head evidence if that review is complete; otherwise the documented merge gate remains blocked.

Maintainers can handle the repository labels separately.

rebelle5868:
Both items from your last review are done.

The verification result now reads 157 tests. The command changed too: the earlier `|| break` form exits `0` when a test fails, so it could report a false pass.

The documentation-writer record is completed and names what the review actually found. It ran eight rounds, each one running every command both documents give. It was not cosmetic — it corrected the walkthrough recording two model turns while both documents said one, a claim that the seed memory drives the intent gate in the fixture path when it does not, two claims about Hermes contradicted by Hermes's own source, two shipped skills reading a path nothing creates, and a `workspace` update hazard stated backwards.

A separate reviewer working from the shell reached two of those findings independently, and additionally ran `hermes profile install --force` and `hermes profile update` against a real runtime — the one claim in this README that no earlier round could check. It holds.

On this head I re-ran the documented verification myself: seven walkthrough steps and exit 0; `"added": 8` then `"added": 0`; 157 tests with `failed=0`, matching discovery; every documented exit code reproduced; all four external links 200; license headers and `git diff --check` clean.

Two follow-ups are recorded in the checklist rather than fixed here: a few British spellings remain in `.py` comments, and `preferences.collect` joins three tables — fine at this scale, and its docstring now says why that does not travel to a store whose tables are id-linked silos.

apurvvkumaria (approved):
Final maintainer review complete. The remaining verification-command and documentation-review issues are corrected. The documented suite reports 157 passing tests, the repository SPDX and diff checks pass, all required GitHub checks pass, and no review conversations remain unresolved.

PR #125, opened 2026-08-20
feat(examples): add the installer, scheduled jobs, and selectors for memory-driven chief-of-staff

rebelle5868:
## Related Issue

Part of #122. This is the second of the three sequential phases agreed there; it does not close the issue.

## Description

Adds the installer, the scheduled jobs, and the two selector scripts that gate them. The store from #123 runs by hand; this makes it run on a schedule without paying for the ticks where there is nothing to do.

`scripts/install.sh` installs the distribution, transfers three named non-secret model settings through the Hermes CLI, verifies that they were applied, requires a credential on the target profile unless keyless operation is explicitly selected, and calls `scripts/register-jobs.sh`, which registers five jobs through the cron CLI: intake every thirty minutes, review every six hours, and three daily maintenance jobs.

The two message-facing jobs run a selector script first, then one agent turn over its output. When the selector finds nothing it prints `{"wakeAgent": false}` as its last line and Hermes skips the agent entirely — no model call, no delivery. That last-line detail is the whole mechanism, and the test asserts it the way `cron/scheduler.py` parses it rather than looking for the string anywhere in the output.

Nothing here reaches a network. The connectors remain Phase 3; the intake selector looks for them, reports each as `"absent": true`, and carries on.

## Defects found and fixed before this was pushed

Four of these were in the implementation this phase was ported from; one I introduced while writing it.

| Defect | Consequence had it shipped |
| --- | --- |
| `register-jobs.sh` looked jobs up with `hermes cron list --json` | That flag does not exist. argparse printed usage, `2>/dev/null` swallowed it, and the lookup returned empty every time — so a script whose header says "Idempotent" created another copy of all five jobs on every run |
| The wake-gate test asserted the gate appeared *somewhere* in stdout | Hermes reads only the last non-empty line. A gate printed before the payload would be ignored and every idle tick would wake the model, with the test still green |
| `awk '/^Path:/ {print $2}'` in both scripts | Truncated any profile path containing a space |
| The teardown line named `hermes profile remove` | No such subcommand; it is `delete`. Written by me, caught by checking the CLI |
| "Credentials are deliberately not copied" | The script copies the user's whole `config.yaml`, and two documented keys there — `api_key` and `sudo_password` — hold secrets if set. It copies `config.yaml` and never `.env` or `auth.json`; it cannot promise more |
| The scheduling section offered macOS as a supported platform | Every shipped skill declares `platforms: [linux]`. On macOS the jobs fire, the model is called, and no skill loads — a scheduled expense with no assistant. The scheduled path is now documented as Linux or WSL only |
| `gateway start` was given as the way to start the gateway | It exits 1 on Linux until `gateway install` has run, and on WSL without systemd. Both steps are documented now, with `gateway run` for the foreground case |
| `hermes -p <profile> model <name>` in the installer's remediation | `hermes model` takes no positional argument. It is `config set model <name>` |
| The friendly error in both scripts was unreachable | Under `set -euo pipefail`, `hermes profile show` exiting 1 aborted at the assignment. The script exited 1 silently instead of saying "Install the profile first" |
| `cron list \| head -20` above "remove one per job above" | `cron list` is 51 lines for five jobs, so three of the five ids were never shown |

The idempotency fix reads the job store the scheduler itself writes, because no cron subcommand emits machine-readable output. Writes still go through the CLI.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — `All 413 checked source files have SPDX headers.`
- [ ] `python3 scripts/check_label_taxonomy.py --check` — not applicable; this branch changes no governance metadata.
- [x] `bash -n` on both shell scripts — clean.
- [x] Relevant example checks:
  - `fail=0; for t in tests/*.py; do python3 "$t" || fail=1; done; echo "failed=$fail"` from `profile/scripts` — 223 tests, every file `OK`, `failed=0`. Discovery reports the same 223.
  - `HERMES_HOME=$(mktemp -d)` then `python3 profile/scripts/walkthrough.py --fixtures fixtures` — seven steps, exit 0.

**Runtime validation.** Installed into a scratch `HERMES_HOME` and exercised against a real Hermes 0.20.0:

A live scheduled turn through a reachable model was not independently reproduced during maintainer review. The selector output, decision-envelope application, and resulting store mutation are covered by the deterministic tests and fixture walkthrough.

- `hermes profile install ./profile --name p2test --force --yes` — installed.
- `register-jobs.sh` run twice — first run `creating` five, second run `updating` five, store holds five jobs, not ten.
- `hermes profile update p2test` — the five jobs and their history survive, which is what "the distribution does not own `cron`" means.
- `hermes cron status` — reports `Gateway is not running — cron jobs will NOT fire`, which is the README's "registering is not starting".

A hundred new tests cover: selector output and bounds, the gate under the scheduler's own parsing rule, the manifest never claiming `cron` or `workspace`, the registration script's lookup and write paths, the documented schedule matching the script's arguments, what survives a reboot, and every `hermes` command the scripts *and the README* name being real — four command groups, plus the flags-only case that `hermes model` is. The four review findings added their own: a collector that exits non-zero still waking the model, both scripts refusing a non-Linux kernel before anything is installed or registered, `INTAKE_SLICE` and `REVIEW_BATCH` rejecting values outside their range rather than passing a negative `LIMIT` through to SQLite, the installer carrying named settings and never a file or a credential, with its model check and its credential check both landing before job registration, the README naming no environment variable the script ignores, and the test count the README quotes being asserted against the suite rather than maintained by hand. The second round added behavioural coverage of the installer — `hermes` stubbed on `PATH`, asserting that a failed or non-sticking transfer aborts and schedules nothing while a clean one still reaches registration — and of the prompt boundary. The final round verifies that secret-shaped collector diagnostics reach neither the model prompt nor persistent scheduler logs; only sanitized failure metadata is retained.

- Result: `docs-updated`
- Reviewer: Xuan Wu (@rebelle5868), author review of the changed documentation. Maintainer review remains required before approval.
- Evidence or justification: the changed text was reviewed against `WRITING.md`, the controlled-word list, and CONTRIBUTING's twelve items, on Hermes 0.20.0 with an isolated `HERMES_HOME`. The review reproduced rather than read: the `wakeAgent` gate against `cron/scheduler.py::_parse_wake_gate` and both selectors under work-present and no-work conditions; registration idempotency across three runs, yielding the same five job ids; `hermes profile update` preserving all five jobs and `cron/executions.db`; the five-row job table against the live `cron/jobs.json`; the absent-connector path; and, on a NemoClaw runtime, that the installer transfers three named model settings and leaves the installed profile with zero occurrences of `api_key`. It returned `changes-required` with five blocking findings, all fixed here and listed in the table above, and twelve smaller ones, also fixed. Its most consequential finding was the platform gate: every shipped skill declares `platforms: [linux]`, and `skill_matches_platform_list(['linux'])` returns `False` on macOS, so a macOS reader following the scheduling section would have got a job that wakes the model every half-hour with no skill attached. The README now states that the scheduled path is Linux or WSL only, and says why. Changed documentation paths: `examples/recipes/nvidia/memory-driven-chief-of-staff/README.md`.

## Release And Compliance

- [x] No secrets or credentials are included. The installer copies no file at all: it transfers three named model settings — `model.default`, `model.provider`, `model.base_url` — through `hermes config set`, so nothing it writes can carry a key. Checked on a NemoClaw runtime, where the generated `config.yaml` holds an inline `api_key` in the same `model:` block: the installed profile came out with zero occurrences of `api_key`. The credential is not inherited either — a profile without one sends the placeholder `no-key-required` — so the installer requires `model.api_key` on the target profile and exits before registering any job without it.

### DIFF
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
@@ -63,6 +63,19 @@ def collect() -> tuple[dict[str, object], bool]:
     working, a tick that skipped the agent, and nothing anywhere saying so,
     every half hour.
 
+    What a failure reports is deliberately narrow. This function's stdout is
+    the scheduled agent's prompt, and a collector's stderr is arbitrary text
+    from a subprocess that talks to a mail or chat API: a traceback carrying a
+    bearer token, a request URL with a signature in the query string, a
+    response body full of someone's messages. Truncating that to a couple of
+    hundred characters bounds its length, not its content — the first two
+    hundred characters of a traceback are exactly where the request line is.
+
+    So the prompt gets a stable error class and an exit code, which is all the
+    agent can act on anyway, and the operator gets the detail on this
+    process's own stderr, where cron writes it to a local log instead of
+    sending it to an inference provider.
+
     Returns the per-collector results and a flag: true if any of them failed.
     """
     results: dict[str, object] = {}
@@ -74,19 +87,31 @@ def collect() -> tuple[dict[str, object], bool]:
             continue
         proc = subprocess.run([sys.executable, str(path)],
                               capture_output=True, text=True)
-        detail = (proc.stderr or "").strip()[:200]
         if proc.returncode != 0:
             failed = True
             results[script] = {"failed": True, "exit_code": proc.returncode,
-                               "stderr": detail}
+                               "error_class": "nonzero_exit"}
+            _report(script, proc.stderr)
             continue
         try:
             results[script] = json.loads(proc.stdout or "{}")
-        except json.JSONDecodeError as exc:
+        except json.JSONDecodeError:
             failed = True
-            results[script] = {"failed": True, "error": f"unreadable output: {exc}",
-                               "stderr": detail}
+            results[script] = {"failed": True, "exit_code": proc.returncode,
+                               "error_class": "unreadable_output"}
+            _report(script, proc.stderr)
     return results, failed
+
+
+def _report(script: str, stderr: str | None) -> None:
+    """Put a failing collector's own output where an operator can read it.
+
+    Not on stdout: that is the agent's prompt. This goes to stderr, which the
+    scheduler captures into the job's local log.
+    """
+    detail = (stderr or "").strip()
+    if detail:
+        print(f"{script}: {detail}", file=sys.stderr)
 
 
 def main() -> int:
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
@@ -63,18 +63,24 @@ def collect() -> tuple[dict[str, object], bool]:
     working, a tick that skipped the agent, and nothing anywhere saying so,
     every half hour.
 
-    What a failure reports is deliberately narrow. This function's stdout is
-    the scheduled agent's prompt, and a collector's stderr is arbitrary text
-    from a subprocess that talks to a mail or chat API: a traceback carrying a
-    bearer token, a request URL with a signature in the query string, a
-    response body full of someone's messages. Truncating that to a couple of
-    hundred characters bounds its length, not its content — the first two
-    hundred characters of a traceback are exactly where the request line is.
+    A failing collector's own output is never repeated anywhere. It is
+    arbitrary text from a subprocess that talks to a mail or chat API: a
+    traceback carrying a bearer token, a request URL with a signature in the
+    query string, a response body full of someone's messages.
 
-    So the prompt gets a stable error class and an exit code, which is all the
-    agent can act on anyway, and the operator gets the detail on this
-    process's own stderr, where cron writes it to a local log instead of
-    sending it to an inference provider.
+    Two places wanted it and neither may have it. This function's stdout is
+    the scheduled agent's prompt, so putting it there sends it to an inference
+    provider. This process's stderr is captured by the scheduler into the job
+    log, so putting it there turns something transient into something stored —
+    a log that outlives the token in it, on disk, for as long as logs are kept.
+    Truncating bounds length rather than content; the first two hundred
+    characters of a traceback are exactly where the request line is.
+
+    So both get the same sanitized triple — collector name, exit code, stable
+    error class — which is what the agent can act on and what tells an operator
+    which collector to go and run by hand. Running it directly is how you read
+    what it actually said, and that output goes to a terminal rather than a
+    file.
 
     Returns the per-collector results and a flag: true if any of them failed.
     """
@@ -91,7 +97,7 @@ def collect() -> tuple[dict[str, object], bool]:
             failed = True
             results[script] = {"failed": True, "exit_code": proc.returncode,
                                "error_class": "nonzero_exit"}
-            _report(script, proc.stderr)
+            _report(script, proc.returncode, "nonzero_exit")
             continue
         try:
             results[script] = json.loads(proc.stdout or "{}")
@@ -99,19 +105,21 @@ def collect() -> tuple[dict[str, object], bool]:
             failed = True
             results[script] = {"failed": True, "exit_code": proc.returncode,
                                "error_class": "unreadable_output"}
-            _report(script, proc.stderr)
+            _report(script, proc.returncode, "unreadable_output")
     return results, failed
 
 
-def _report(script: str, stderr: str | None) -> None:
-    """Put a failing collector's own output where an operator can read it.
+def _report(script: str, exit_code: int, error_class: str) -> None:
+    """Say which collector failed and how, without quoting it.
 
-    Not on stdout: that is the agent's prompt. This goes to stderr, which the
-    scheduler captures into the job's local log.
+    This goes to stderr, which the scheduler stores in the job log — so it
+    carries only what is safe to keep: the collector's name, its exit code, and
+    a stable class. The collector's own text is deliberately dropped rather
+    than redacted, because a pattern-matching redactor cannot promise it caught
+    everything and a log is the wrong place to find out.
     """
-    detail = (stderr or "").strip()
-    if detail:
-        print(f"{script}: {detail}", file=sys.stderr)
+    print(f"{script}: failed (exit {exit_code}, {error_class}). "
+          f"Run it directly to see its output.", file=sys.stderr)
 
 
 def main() -> int:
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
@@ -19,18 +19,54 @@ from pathlib import Path
 
 from _db import ensure_store, write_txn
 
+def bounded_int(name: str, default: int, *, maximum: int) -> int:
+    """Read a positive, bounded integer from the environment.
+
+    `LIMIT` takes whatever it is given, and SQLite reads a negative one as no
+    limit at all — so `INTAKE_SLICE=-1` handed the model every pending row in
+    the store, silently defeating the bound this recipe is built on. Zero is
+    just as wrong in the other direction: the job wakes, reports work, and
+    offers nothing. Malformed text used to raise during import, before any
+    error message could explain which variable was at fault.
+    """
+    raw = os.environ.get(name)
+    if raw is None or raw == "":
+        return default
+    try:
+        value = int(raw)
+    except ValueError:
+        raise SystemExit(
+            f"{name} must be a whole number between 1 and {maximum}; got {raw!r}")
+    if value < 1 or value > maximum:
+        raise SystemExit(
+            f"{name} must be between 1 and {maximum}; got {value}")
+    return value
+
+
 HERE = Path(__file__).resolve().parent
-SLICE = int(os.environ.get("INTAKE_SLICE", "25"))
+MAX_SLICE = 200
+SLICE = None  # resolved in main(); see bounded_int
 
 
-def collect() -> dict[str, object]:
-    """Run whichever collectors are present and configured.
+def collect() -> tuple[dict[str, object], bool]:
+    """Run whichever collectors are present, and report whether any failed.
 
     A collector that is absent is not an error. The store and the schedule ship
     before the connectors do, and a reader who never connects a source still
     gets a working intake pass over whatever is already in the store.
+
+    A collector that *fails* is a different thing, and it must be visible. The
+    exit code is what says so: a connector whose credential expired writes to
+    stderr and exits non-zero while printing nothing, and `json.loads("" or
+    "{}")` turns that into an empty success. Combined with the idle gate below
+    it produced the worst failure this design can have — a token that stopped
+    working, a tick that skipped the agent, and nothing anywhere saying so,
+    every half hour.
+
+    Returns the per-collector results and a flag: true if any of them failed.
     """
     results: dict[str, object] = {}
+    failed = False
     for script in ("ingest_graph.py", "ingest_slack.py"):
         path = HERE / script
         if not path.exists():
@@ -38,18 +74,26 @@ def collect() -> dict[str, object]:
             continue
         proc = subprocess.run([sys.executable, str(path)],
                               capture_output=True, text=True)
+        detail = (proc.stderr or "").strip()[:200]
+        if proc.returncode != 0:
+            failed = True
+            results[script] = {"failed": True, "exit_code": proc.returncode,
+                               "stderr": detail}
+            continue
         try:
             results[script] = json.loads(proc.stdout or "{}")
-        except json.JSONDecodeError:
-            # A collector that fails must not take the whole tick down; the
-            # slice below still has yesterday's unjudged items to work on.
-            results[script] = {"error": (proc.stderr or "").strip()[:200]}
-    return results
+        except json.JSONDecodeError as exc:
+            failed = True
+            results[script] = {"failed": True, "error": f"unreadable output: {exc}",
+                               "stderr": detail}
+    return results, failed
 
 
 def main() -> int:
+    global SLICE
+    SLICE = bounded_int("INTAKE_SLICE", 25, maximum=MAX_SLICE)
     ensure_store()
-    collected = collect()
+    collected, collector_failed = collect()
 
     with write_txn() as conn:
         rows = conn.execute(
@@ -75,7 +119,9 @@ def main() -> int:
     }
     print(json.dumps(payload, ensure_ascii=False, indent=2))
 
-    if not rows:
+    # Wake on a collector failure even with nothing to judge. The gate exists
+    # to make an idle tick free, not to make a broken one quiet.
+    if not rows and not collector_failed:
         print(json.dumps({"wakeAgent": False}))
     return 0
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
@@ -1,0 +1,84 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Cron pre-step for the intake pass.
+
+Runs the collectors, then hands the agent a slice of unjudged items. When
+there is nothing new it emits `{"wakeAgent": false}`, which Hermes treats as a
+signal to skip the agent turn entirely — so a quiet half-hour costs no tokens
+at all rather than paying a model call to be told there is no work.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+from _db import ensure_store, write_txn
+
+HERE = Path(__file__).resolve().parent
+SLICE = int(os.environ.get("INTAKE_SLICE", "25"))
+
+
+def collect() -> dict[str, object]:
+    """Run whichever collectors are present and configured.
+
+    A collector that is absent is not an error. The store and the schedule ship
+    before the connectors do, and a reader who never connects a source still
+    gets a working intake pass over whatever is already in the store.
+    """
+    results: dict[str, object] = {}
+    for script in ("ingest_graph.py", "ingest_slack.py"):
+        path = HERE / script
+        if not path.exists():
+            results[script] = {"absent": True}
+            continue
+        proc = subprocess.run([sys.executable, str(path)],
+                              capture_output=True, text=True)
+        try:
+            results[script] = json.loads(proc.stdout or "{}")
+        except json.JSONDecodeError:
+            # A collector that fails must not take the whole tick down; the
+            # slice below still has yesterday's unjudged items to work on.
+            results[script] = {"error": (proc.stderr or "").strip()[:200]}
+    return results
+
+
+def main() -> int:
+    ensure_store()
+    collected = collect()
+
+    with write_txn() as conn:
+        rows = conn.execute(
+            "SELECT source_id, source, scope, event_at, sender, subject, body,"
+            "       addressing, unread"
+            "  FROM items WHERE state='pending'"
+            " ORDER BY event_at LIMIT ?", (SLICE,)).fetchall()
+        open_rows = conn.execute(
+            "SELECT o.source_id, o.title, o.status FROM obligations o"
+            " WHERE o.status='open' ORDER BY o.global_rank LIMIT 40").fetchall()
+        recent_closed = conn.execute(
+            "SELECT o.title, o.status FROM obligations o"
+            " WHERE o.status IN ('done','ignored')"
+            " ORDER BY o.updated_at DESC LIMIT 20").fetchall()
+
+    cols = ("source_id", "source", "scope", "event_at", "sender", "subject",
+            "body", "addressing", "unread")
+    payload = {
+        "collected": collected,
+        "slice": [dict(zip(cols, r)) for r in rows],
+        "open_obligations": [{"source_id": r[0], "title": r[1]} for r in open_rows],
+        "recently_resolved": [{"title": r[0], "status": r[1]} for r in recent_closed],
+    }
+    print(json.dumps(payload, ensure_ascii=False, indent=2))
+
+    if not rows:
+        print(json.dumps({"wakeAgent": False}))
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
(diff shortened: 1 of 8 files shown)


rebelle5868:
Validated on a real Linux runtime, which turned up one defect the machine that produced the branch cannot reproduce.

**Environment.** A NemoClaw sandbox running Hermes 0.19.0 — the floor `profile/distribution.yaml` declares — on Ubuntu 24.04. Everything below ran inside that container against a scratch `HERMES_HOME`.

**What the platform gate actually does.** The README says the scheduled path is Linux or WSL only, because every shipped skill declares `platforms: [linux]`. Both directions are now measured rather than inferred. On macOS `skill_matches_platform_list(['linux'])` returns `False`, and all five skills fail to load. In the sandbox:

```
platform: linux
  platforms=['linux']  -> True

  inbound-judging        loaded=True
  obligation-review      loaded=True
  memory-repair          loaded=True
  memory-consolidation   loaded=True
  preference-update      loaded=True
```

**The wake gate, judged by the scheduler rather than by the test.** I fed each selector's real output to Hermes's own `cron.scheduler._parse_wake_gate`:

```
empty store        last line: {"wakeAgent": false}   _parse_wake_gate -> False
8 fixtures loaded  last line: }                      _parse_wake_gate -> True
```

`hermes profile install` and `register-jobs.sh` both behaved as documented; five jobs, created once.

**The defect.** The first archive I shipped over was packed on macOS, so it carried an AppleDouble sidecar beside each file. `._install.sh` matches `*.sh` and is binary, and two test files raised `UnicodeDecodeError` unpacking on Linux — while the same suite was green on the Mac that produced the archive.

Phase 1 had already solved this for memory pages: `load_fixtures` filters `._*` when it copies, and `memory_check._pages` skips dotfiles, with a comment noting that a macOS archive carries them along. The scans added for this phase did not inherit that, and neither did five older globs in `test_durability.py`. They share one helper now. `walkthrough.py` was exposed in a quieter way — `next((root / "people").glob("*.md"), None)` had the sidecar in its candidate set and returned a real page by filesystem order rather than by rule.

Fixed in `b1ae25b`, with three tests: the helper keeps originals and drops sidecars, every scan reads its real files without raising, and an unguarded glob does raise. Re-verified in the sandbox — green with a clean archive, and still green after a sidecar is dropped back in. 185 tests, direct execution and discovery agreeing; the PR description and the README now say 185.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/install.sh
```diff
@@ -0,0 +1,74 @@
...
+PROFILE_HOME="$(hermes profile show "$PROFILE" 2>/dev/null \
+  | sed -n 's/^Path:[[:space:]]*//p' || true)"
+if [[ -z "$PROFILE_HOME" ]]; then
+  echo "Could not resolve the profile home for '$PROFILE'." >&2
+  exit 1
+fi
+
+echo "2/3  Inheriting the runtime's model configuration"
+if [[ -f "$SOURCE_PROFILE_CONFIG" ]]; then
+  cp "$SOURCE_PROFILE_CONFIG" "$PROFILE_HOME/config.yaml"
```
apurvvkumaria: Hermes cron jobs use the target profile’s own `config.yaml`, `.env`, and skills. Copying only `config.yaml` leaves a normal `.env` API key behind, so the scheduled profile cannot authenticate; if the source config contains an inline API key or plaintext `sudo_password`, this copies that secret instead. Please do not copy the whole config. Transfer only the required non-secret model/provider settings through the supported CLI, require credential setup in the target profile, verify model/provider resolution, and abort before job registration when the profile is not runnable. The README’s `No credentials of any kind` statement must be scoped or corrected too.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/install.sh
```diff
@@ -0,0 +1,74 @@
...
+PROFILE="${PROFILE_NAME:-memory-driven-chief-of-staff}"
+SOURCE_PROFILE_CONFIG="${SOURCE_PROFILE_CONFIG:-$HOME/.hermes/config.yaml}"
+
+command -v hermes >/dev/null 2>&1 || {
+  echo "hermes is not on PATH." >&2
+  exit 1
+}
+
+echo "1/3  Installing the profile distribution"
+hermes profile install "$RECIPE_ROOT/profile" --name "$PROFILE" --force --yes
```
apurvvkumaria, line 53: The README correctly explains that this scheduled path is known not to work on macOS, but the installer still performs the installation and registers jobs there. That produces the exact model-without-skill calls the documentation warns about. Add a Linux/WSL guard before the first mutation, protect direct execution of `register-jobs.sh` as well, and test Darwin rejection plus Linux acceptance.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
```diff
@@ -0,0 +1,84 @@
...
+    results: dict[str, object] = {}
+    for script in ("ingest_graph.py", "ingest_slack.py"):
+        path = HERE / script
+        if not path.exists():
+            results[script] = {"absent": True}
+            continue
+        proc = subprocess.run([sys.executable, str(path)],
+                              capture_output=True, text=True)
+        try:
+            results[script] = json.loads(proc.stdout or "{}")
```
apurvvkumaria, line 103: Check `proc.returncode` before accepting stdout. I reproduced a collector exiting 1 with empty stdout and `connector failed` on stderr being reported as `{}`. Even parse errors that reach the `error` branch become invisible when the store has no pending rows because `wakeAgent=false` suppresses the entire tick. Preserve the failure while continuing to process existing pending rows, but do not emit the idle gate when any collector failed. Add coverage for nonzero empty output, nonzero valid JSON, and a failed collector with no pending rows.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
```diff
@@ -0,0 +1,84 @@
...
+import json
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+from _db import ensure_store, write_txn
+
+HERE = Path(__file__).resolve().parent
+SLICE = int(os.environ.get("INTAKE_SLICE", "25"))
```
apurvvkumaria: Validate this override before passing it to SQLite. `LIMIT -1` means unlimited, so `INTAKE_SLICE=-1` sends every pending message to the model; `REVIEW_BATCH` has the same behavior, and malformed text currently crashes during import. Require a positive bounded integer for both selectors and add negative, zero, and malformed-value tests.

apurvvkumaria (changes requested):
Thank you for the AppleDouble sidecar fix and the added Linux validation. I reviewed `b1ae25b` and reran both documented test methods: all 185 tests pass. Shell syntax, SPDX headers, diff checks, DCO, and governance checks also pass.

The four remaining blockers are attached inline. Before approval, please also:

- demonstrate one agent-backed scheduled intake run with pending fixture data and confirm that the writer changes the store; the current runtime validation stops at installation, registration, and wake-gate parsing;
- identify the documentation reviewer in the PR description; and
- update `Twenty-five new tests` to 28, accounting for the three sidecar tests.

Maintainers can apply the repository labels after the technical findings are resolved.

rebelle5868:
Thanks — all four are fixed in `bf565f8`, and the fourth turned out to be worse than the finding said.

**Collector failure visibility.** `select_intake.py` decided the wake gate purely on the count of pending rows, so a collector that exited non-zero left nothing pending and the tick printed `wakeAgent: false` — a broken run and an empty mailbox were indistinguishable. The gate now wakes on a collector failure even with nothing to judge.

**Platform guard.** Both scripts refuse to run outside Linux, before anything is installed or registered. WSL reports Linux and is supported.

**Bounded slice.** `INTAKE_SLICE` and `REVIEW_BATCH` went straight into `LIMIT`, and SQLite reads a negative limit as no limit, so `INTAKE_SLICE=-1` handed the model every pending row. Both now parse through a bounded reader that rejects anything outside its range.

**Config copy.** Replaced with three named settings through `hermes config set`, plus a resolution check that exits before job registration. Checking this on a NemoClaw runtime turned up the concrete version of the risk you flagged: the generated `config.yaml` carries `default`, `provider`, `base_url` and `api_key` together in the `model:` block, so the old wholesale copy wrote the inference key into a second file. Carrying only the three settings loses nothing — a profile whose own `model.api_key` is unset still sent an authenticated request, the credential resolving from the config it inherits, which I confirmed by reading the request dump of a cron-driven turn.

End to end on Linux, on that runtime: the installer aborted on a home with no resolvable model; after one was set it installed the profile and registered all five jobs; the new profile's `config.yaml` held the three settings and zero occurrences of `api_key`; `hermes cron run` fired the intake job and the pre-step selected the fixture batch and woke the agent with it; the writer took the store from 0 to 6 obligations with 2 skipped and the documented tier split; and the pre-step on the now-judged store printed `wakeAgent: false`.

One step I could not complete there is the model's own turn — the managed endpoint is not routable from that shell, so the agent's request failed to connect. The link between the model replying and the writer running is still covered by tests and the walkthrough rather than by a live run.

The README paragraphs describing the copy and the `SOURCE_PROFILE_CONFIG` override are rewritten, and its test count is now asserted by the suite rather than maintained by hand — it had drifted three times before anything checked it. 210 tests, green on Linux, direct execution and discovery agreeing.

rebelle5868:
Correction to my previous comment: the claim that a profile carrying only the three model settings still authenticates was wrong, and the evidence I offered for it did not support it. I read a request dump, saw a populated `Authorization` header, and concluded the credential had resolved from the inherited config.

It had not. Pointing a scratch profile at a local endpoint and reading what actually arrived shows the header carries the literal string `no-key-required` — fifteen bytes of placeholder against a configured key of twenty-six, different hashes. `hermes -p <profile> config show` confirms it from the other side: the `Model` block of a profile installed this way has no `api_key`, while the profile it was installed alongside does.

So your original instruction stands and I have implemented it in `933986d`. The installer now requires `model.api_key` on the target profile and exits before `register-jobs.sh` when it is missing, alongside the model check. `ALLOW_NO_API_KEY=1` opts out for endpoints that genuinely need none. Verified on Linux that the gate stops before registration and leaves no `cron/jobs.json` behind. The README passages and installer comments that repeated the wrong claim are corrected, including the honest limit: the installer proves a key is present, not that it is valid.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/install.sh
rebelle5868: Done in `bf565f8` and `933986d`, all five parts.

The copy is gone: the installer transfers `model.default`, `model.provider` and `model.base_url` through `hermes config set`. Your inline-secret case turned out to be the common one rather than the edge case — on the NemoClaw runtime I checked against, the generated `config.yaml` carries `api_key` in the same `model:` block as those three, so the old copy wrote the inference key into a second file on every install. The installed profile now comes out with zero occurrences of `api_key`.

Credential setup in the target profile is now required, and I owe you a correction on why it took a second commit. I first argued it was unnecessary because the profile inherits the credential, citing a populated `Authorization` header in a request dump. That was wrong. Reading what actually reaches an endpoint shows the header carries the literal placeholder `no-key-required`, not the configured key; `config show` agrees, listing no `api_key` in the profile's `Model` block while the neighbouring profile has one. The installer now requires `model.api_key` and exits before `register-jobs.sh` without it, with `ALLOW_NO_API_KEY=1` as an explicit opt-out for keyless endpoints. It reads the value and never echoes it.

Model and provider resolution is verified the same way, and both checks land before registration — verified on Linux that a profile missing either ends the run with no `cron/jobs.json` written at all. The README's `No credentials of any kind` line is scoped to the fixture path, and the paragraph describing the copy and `SOURCE_PROFILE_CONFIG` is rewritten; that variable no longer exists. The stated limit is that the installer proves a key is present, not that it is valid.

Coverage: `TestTheInstallerCarriesSettingsNotSecrets` — no file copy, the three settings through the CLI, nothing credential-shaped transferred, both checks preceding registration, an unresolvable model and a missing credential each exiting non-zero, the opt-out existing, and the credential never printed.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/install.sh
rebelle5868: Done. `require_linux()` is called before the first mutation in both scripts: in `install.sh` at line 42, ahead of `hermes profile install` at line 54 and `register-jobs.sh` at line 135; `register-jobs.sh` carries its own copy at line 38, so running it directly is guarded too. WSL reports `Linux` and is accepted.

Coverage in `TestTheInstallerRefusesTheWrongPlatform`, with `uname` stubbed: Darwin rejected by both scripts, Linux accepted and reaching the next check, the refusal naming the fixture path that does work on any platform, and an assertion that the guard's position precedes every install and registration call — so it cannot drift below a mutation in a later edit.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
rebelle5868: Done. `collect()` now returns `(results, failed)`. A non-zero exit records `{"failed": true, "exit_code": …, "stderr": …}` with the first 200 characters of stderr, and a clean exit whose stdout will not parse records the parse error the same way; both set the failure flag. Pending rows are still selected and offered in the same tick, so a broken collector does not stop work already in the store. The idle gate is emitted only when there are no rows *and* no collector failed — it exists to make an idle tick free, not to make a broken one quiet.

Coverage in `TestACollectorFailureIsVisible`, matching the three cases you named: non-zero exit with empty stdout, non-zero exit with valid JSON on stdout, and a failure with no pending rows still waking the agent. Two more hold the other side: a healthy collector with no rows still gates, and a failure does not stop pending rows being offered.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
rebelle5868: Done. Both selectors read their override through a shared `bounded_int()` that requires a whole number in range and exits with a message naming the variable, so nothing reaches `LIMIT` unvalidated. `INTAKE_SLICE` defaults to 25 and `REVIEW_BATCH` to 15, both capped at 200. The value is resolved in `main()` rather than at import, so a malformed value produces that message instead of a traceback while the module loads.

Coverage in `TestTheSliceBoundCannotBeDefeated`: negative, zero, malformed text, above the ceiling, the default, and a valid override still taking effect. You named `REVIEW_BATCH` alongside `INTAKE_SLICE` and only the negative case ran against both, so `a7c6b74` parameterises every case over both selectors. The suite is at 213 tests as of `933986d`.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/install.sh
```diff
@@ -0,0 +1,145 @@
...
+# config really does put one there — on the NemoClaw runtime this recipe was
+# checked against, `model:` carries `default`, `provider`, `base_url` and
+# `api_key` together. A wholesale copy writes that key into a second file.
+#
+# What the three settings do not carry is a credential, and nothing else
+# supplies one either — see the check below.
+for key in model.default model.provider model.base_url; do
+  value="$(hermes config get "$key" 2>/dev/null | head -1 || true)"
+  [[ -z "$value" || "$value" == *"not set"* ]] && continue
+  hermes -p "$PROFILE" config set "$key" "$value" >/dev/null 2>&1 \
```
apurvvkumaria: `set -e` does not exit when a command fails as the left operand of `&&`. I reproduced `config set model.provider` and `config set model.base_url` returning 1: this loop silently skipped both, the model/key checks passed, and the installer registered all five jobs and exited 0. Fail closed when any non-empty source setting cannot be applied, then validate the resulting provider/route before registration. This is the remaining part of the model/provider-resolution requirement.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
```diff
@@ -0,0 +1,130 @@
...
+        path = HERE / script
+        if not path.exists():
+            results[script] = {"absent": True}
+            continue
+        proc = subprocess.run([sys.executable, str(path)],
+                              capture_output=True, text=True)
+        detail = (proc.stderr or "").strip()[:200]
+        if proc.returncode != 0:
+            failed = True
+            results[script] = {"failed": True, "exit_code": proc.returncode,
```
apurvvkumaria, line 98: This `stderr` value becomes part of the JSON written to stdout, and that stdout is the scheduled agent prompt. Collector diagnostics can contain credential values, request URLs, private identifiers, or response bodies, so truncation is not sufficient redaction. Keep the prompt payload generic (for example, failure state, exit code, and a stable error class), write only sanitized detail to local stderr or a protected log, and add a test proving secret-shaped collector stderr never appears in selector stdout.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/tests/test_selectors.py
```diff
@@ -0,0 +1,746 @@
...
+
+
+class TestTheInstallerCarriesSettingsNotSecrets(unittest.TestCase):
+    """The installer used to copy `~/.hermes/config.yaml` into the new profile.
+
+    That file's own `model:` block is documented to hold an inline `api_key`,
+    and a generated config really does put one there, so the copy duplicated a
+    credential into a second file while the README told the reader no
+    credentials were involved. It also bought nothing: a profile with no key of
+    its own authenticates through the config it inherits.
```
apurvvkumaria: This sentence still states the opposite of the corrected behavior established in `933986d`: a target profile without its own key sends `no-key-required`. Please update the class docstring so the test documentation agrees with the installer, README, and the later credential test.

apurvvkumaria (changes requested):
Thank you for addressing the original four findings and for correcting the credential analysis. I reviewed `933986d` and reran both documented test methods: all 213 tests pass. Shell syntax, SPDX, diff, DCO, and governance checks also pass.

Two remaining runtime-safety blockers and one stale statement are attached inline. Before approval, please also complete the previously requested agent-backed scheduled intake validation with a reachable model: load pending fixtures, trigger the registered intake job, confirm the model produces the envelope, and show that the scheduled writer changes the store. The current evidence explicitly says the model request could not connect, so the walkthrough and direct writer validation do not exercise that boundary.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/install.sh
rebelle5868: You are right, and I confirmed the mechanism before fixing it: `set -euo pipefail; false && echo x` does not exit. Fixed in `9de690b`.

Every transfer now aborts the run on failure rather than being skipped. Each carried setting is then read back off the target profile and compared, because an exit status is not proof the value landed — a `config set` that returns 0 without writing is caught there. Both happen before either runnability check, so a half-configured profile ends the run with nothing scheduled.

The more useful finding underneath yours is that none of my installer tests could have caught this: every one of them read the script and matched patterns in it, which cannot see a swallowed exit status. Four new tests stub `hermes` on `PATH` and assert on what the installer does — a failed transfer aborts, a failed transfer schedules nothing, a setting that reports success without sticking is caught by the read-back, and a clean transfer still reaches registration. Reverting the fix fails three of them and none of the pattern-matching ones.

They pass `ALLOW_NO_API_KEY` so the credential gate cannot be what stops the run. Without it, my first draft of the schedules-nothing test passed against the broken installer, because a later check happened to abort first.

Verified on Linux against the real CLI as well: a full install carries all three settings, the read-back agrees, and all five jobs register.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
rebelle5868: Agreed, and the truncation was the part I had wrong: it bounds the length, not the content, and the first two hundred characters of a traceback are where the request line is. Fixed in `9de690b`.

The payload now carries `exit_code` and a stable `error_class` — `nonzero_exit` or `unreadable_output` — and nothing quoted from the collector. The detail goes to the selector's own stderr, which the scheduler writes to the job's local log, so an operator still gets it without it having been sent to an inference provider. Unparsable stdout is classed the same way rather than being echoed back into the prompt, which was the other copy of this.

Five tests cover it, including the one you asked for: a collector whose stderr carries a bearer token and a signed URL, asserting none of it reaches stdout, that no raw `stderr` field survives in the payload at all, and that the detail is still present on stderr. Two existing tests asserted the leaking fields directly and were corrected.

The README documents where a failing collector's output goes and why it is split in two.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/tests/test_selectors.py
rebelle5868: Fixed in `9de690b`. The docstring now says what `933986d` established: a profile with no key of its own does not authenticate through the config it inherits, it sends `no-key-required`, so the fix is to require a key on the target profile rather than to copy one.

This is the third time a claim of mine outlived its correction, so I checked the rest rather than just this line: a repo-wide grep for the disproven phrasings returns nothing else. My earlier sweep only covered `README.md` and `install.sh` — the two files I remembered writing it in — which is exactly why this one survived.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_intake.py
```diff
@@ -0,0 +1,155 @@
...
+
+def _report(script: str, stderr: str | None) -> None:
+    """Put a failing collector's own output where an operator can read it.
+
+    Not on stdout: that is the agent's prompt. This goes to stderr, which the
+    scheduler captures into the job's local log.
+    """
+    detail = (stderr or "").strip()
+    if detail:
+        print(f"{script}: {detail}", file=sys.stderr)
```
apurvvkumaria: Moving collector output from stdout to stderr keeps it out of the inference prompt, but `_report()` still writes the raw text unchanged and the scheduler persists it in a job log. As the surrounding documentation notes, that text can contain a bearer token, signed URL, or message body; this turns transient sensitive data into stored log content. Please emit only sanitized metadata such as the collector name, exit code, and stable error class by default. Keep raw diagnostics out of persistent logs, or place them behind an explicit protected diagnostic path with redaction and size limits. Update the test so it proves the secret is absent from both stdout and stderr instead of requiring it on stderr.

apurvvkumaria (changes requested):
The installer fail-closed behavior, setting read-back, prompt redaction, and stale documentation fixes are now correct, and the full 222-test suite passes. One remaining runtime-security blocker is attached inline. Please also complete the previously requested live scheduled-intake validation with a reachable model: load pending fixtures, trigger the registered intake job, confirm that the model produces the decision envelope, and verify that the scheduled writer changes the store. The current evidence still stops short of the model-to-writer boundary.

apurvvkumaria (approved):
Reviewed at `ff56299`. The collector diagnostic path now retains only sanitized failure metadata in both the model prompt and scheduler logs, and the stale guidance has been corrected. All 223 recipe tests, repository governance checks, SPDX validation, DCO, shell syntax, and diff checks pass. The live model-to-writer scheduled turn remains explicitly documented as not independently reproduced; the deterministic selector, envelope, writer, and store-mutation coverage is accepted as the validation boundary for this phase. Approved.

PR #130, opened 2026-08-21
feat(examples): read the user's Slack into the memory-driven chief-of-staff intake

rebelle5868:
## Related Issue

Part three of #122. Phase 1 (#123) landed the store and its tests; phase 2 (#125) landed the installer, the schedule, and the collector seam. This fills that seam.

The determination decision 3 asked for is on #122, along with a correction: my first answer was wrong, and this branch has been reworked onto the rotating credential the approved design calls for. The four data-lifecycle controls #122 lists — retention, export and reset, ingest-time exclusion, and the Slack deletion caveat — landed separately in #137 and are now inherited from `main`. This pull request contains only the connector changes.

## Description

`select_intake.py` already ran `ingest_slack.py` when present and recorded it as absent when not, and that absent path already had tests. This adds the collector, the provider that feeds it a credential, and the setup path.

It reads what the user receives — direct messages, group DMs, and public channels the operator names — and never writes to Slack.

**The gateway holds the credential and refreshes it.** Slack rotates user tokens: the access half lasts twelve hours and each refresh token is single-use. `providers/slack-user.yaml` configures `oauth2-refresh-token` against `oauth.v2.access`; the sandbox receives only an `openshell:resolve:` placeholder that the egress proxy substitutes on the way out. A non-rotating `xoxp-` is refused — not because it would fail, but because a user token that never expires is a permanent key to one person's entire Slack.

**Setup takes the refresh token from `authed_user.refresh_token` by name.** An app with bot scopes returns a second refresh token at the top level belonging to the bot; configuring that one refreshes the bot's credential under a user-token key, reports healthy, and yields a collector that never sees a direct message. Extracting by name makes that impossible even if the manifest later gains bot scopes. The manifest requests no bot scopes, so the app has no "Install to Workspace" button — a user-scopes-only app is authorized through the OAuth URL, which the setup script prints.

**Coverage is bounded and rotates.** Slack documents one `conversations.history` per minute and fifteen messages per response for affected non-Marketplace apps, which is restrictive enough to shape the design rather than be absorbed by retries. A tick spends a request budget, starts where the last one stopped, honours `Retry-After` on `429`, and reports what it could not reach as `incomplete_coverage` rather than letting it look like an absence of messages. Public channels are read only when named in `workspace/slack_channels.json` — the explicit channel list #122 describes. Thread replies are collected through `conversations.replies`, which `conversations.history` does not return.

**Egress is one host, read-only, enforced.** The provider profile declares `slack.com` at `access: read-only` with `enforcement: enforce`, so a write is refused at the boundary before any test runs.

**Real message bodies are gated on encryption at rest**, per decision 5. `setup-slack.sh` inspects the host filesystem under the sandbox's storage and refuses to attach until the prerequisite is confirmed; `docs/encrypted-storage.md` explains why owner-only permissions are a different protection rather than a weaker one, and what to do when the answer is no.

**An unconfigured collector is a state, not a failure.** For one commit it was the latter, and shipping this file would have woken the model every thirty minutes forever for every user who had not connected Slack. A credential that *disappears* is still a failure, told apart by the capability cache the first successful run leaves behind.

Diagnosis rides in the exit code — 2 credential, 3 rate-limited, 4 scope, 1 other — because a failing collector's text is deliberately dropped rather than written to a job log.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — `All 532 checked source files have SPDX headers.`
- [x] `python3 scripts/check_label_taxonomy.py --check` — `Label taxonomy valid: 64 labels`.
- [x] `bash -n` on all four shell scripts — clean.
- [x] Example checks: `python3 -m unittest discover -s tests -p 'test_*.py' -q` from `profile/scripts` — **420 tests**, `OK`, after merging current `main`. The combined suite was run on macOS; the contributor also ran the connector suite on Linux before the merge resolution.
- [x] `openshell provider profile lint --file providers/slack-user.yaml` — `Provider profile lint passed.`

The focused connector tests run the real module against a local stand-in Slack and cover: watermark advance and re-read, a truncated crawl refusing to commit its cursor, one bad conversation not discarding the others, a budget-bounded tick rotating its starting point, thread replies collected and the parent not stored twice, joins and bot posts excluded, the token absent from both streams while present in the header, and each error mapped to its code. Every fix was checked by reverting it and confirming a test goes red.

Verified beyond the test suite, on a NemoClaw sandbox and against live Slack:

- The rotation contract, twice. Against a stub that refuses replayed refresh tokens, four gateway refreshes each carried the token minted by the previous response, status `refreshed`, `expires_at` twelve hours out. Against live Slack, two real refreshes returned top-level `xoxe.xoxp-` credentials with `token_type: user`.
- The whole setup path: profile import, provider create, refresh configure, attach to an **already-running** sandbox, and a cron pre-step receiving `openshell:resolve:env:...` rather than a credential.
- The negative control: configured with the top-level bot refresh token instead of `authed_user.refresh_token`, the gateway refreshes on the bot chain and reports healthy. That is the hazard the by-name extraction exists to prevent.
- The unconfigured path: with no credential set, the shipped collector still lets the idle gate fire.

One result is in the troubleshooting docs rather than the code: `curl https://slack.com/api/api.test` from the sandbox returns `CONNECT tunnel failed, response 403` while the collector's own request to the same URL returns HTTP 200, and swapping `User-Agent` changes neither. The obvious way to check reachability reports the opposite of the truth, so the docs say so and point at the collector instead.

- Result: `docs-updated`
- Reviewer: Xuan Wu (@rebelle5868), author review of the changed documentation. Maintainer review remains required before approval.
- Evidence or justification: reviewed against `WRITING.md`, the controlled-word list, and CONTRIBUTING's twelve items, and re-reviewed after the rework. The first cut drew nineteen findings from an adversarial pass and eight from review here; all are addressed. Several corrected claims rather than defending them — an exit-code table describing behaviour the code did not have, two files calling the collector "absent" when it ships and always runs, and five README passages saying connectors arrive in a later phase and that nothing reaches the network. The claim about which credential names Hermes removes carries the command a reader can run to check it on their own install. Changed documentation paths: `examples/recipes/nvidia/memory-driven-chief-of-staff/README.md`, `.../docs/set-up-slack.md`, `.../docs/encrypted-storage.md`.

## Release And Compliance

- [x] No secrets or credentials are included. The collector reads a placeholder it cannot spend; a test asserts a real token is present in the `Authorization` header and absent from both streams. The setup script reads secrets with `read -s`, passes them through the CLI's documented env-lookup form rather than argv, and unsets them as soon as they are handed over.

### DIFF
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -256,6 +256,7 @@ leave the loop exiting `0`.
 | Writer behavior, audit trail, caps across batches, correction idempotency, correction state transitions, displaced-row audit | `tests/test_apply_decisions.py` |
 | The walkthrough, and its central claims | `tests/test_walkthrough.py` |
 | Selector output, the wake gate, and the scheduler contract | `tests/test_selectors.py` |
+| The Slack collector: watermarks, partial failure, scope probing, and the credential never reaching a stream | `tests/test_ingest_slack.py` |
 
 Four points are worth calling out.
 
@@ -494,8 +495,10 @@ rotation on a Slack app cannot be undone, and nothing here refreshes an
 expiring one, so a rotating token is refused rather than working for an
 afternoon.
 
-Until this is set up the collector is simply absent and the schedule runs over
-whatever is already in the store. That is a supported state, not a broken one.
+Until this is set up the collector still runs — it ships with the recipe — but
+reports `{"unconfigured": true}` and exits zero, so the schedule runs over
+whatever is already in the store and an idle tick still costs nothing. That is
+a supported state, not a broken one.
 
 ### When a collector fails
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -369,10 +369,12 @@ provider takes over from the in-process ticker.
 **The job store is not part of the distribution.** `distribution.yaml` does not
 declare `cron`. An update replaces what it does declare — `SOUL.md`,
 `schema.md`, `skills`, `scripts` and the manifest — and leaves the jobs and
-their run history alone. Measured, not assumed: six jobs
-registered, `hermes profile update` run on Hermes 0.20.0, six jobs still
-there with the same ids. A test asserts
-the manifest never claims `cron` or `workspace`.
+their run history alone. This is worth being deliberate about: `profile
+update` lists `cron/` among the directories it overwrites, and it leaves ours
+alone only because the manifest never claims it. Measured, not assumed: six
+jobs registered, `hermes profile update` run on Hermes 0.19.0, six jobs still
+there with the same ids. A test asserts the manifest never claims `cron` or
+`workspace`.
 
 To undo, remove the jobs individually. Deleting the profile removes its
 `workspace` too, which is where the store and the memory live; removing the
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -244,7 +244,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the eleven files report 403 tests in
+Expected result: every file ends with `OK`, the eleven files report 420 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -244,7 +244,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the ten files report 299 tests in
+Expected result: every file ends with `OK`, the ten files report 303 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -478,6 +478,21 @@ The fixtures were written from scratch. The people, the company, the projects,
 and every message body are invented. Nothing is derived from a real mailbox or
 from an anonymized copy of one. See [`fixtures/README.md`](fixtures/README.md)
 for what each record is a control for.
+
+### Before a connector: encrypted storage
+
+The fixture path stores invented messages. A connector changes that — the store
+then holds real subjects, senders and bodies — so the profile home must sit on
+an encrypted volume before one is attached. Owner-only permissions are not
+encryption: they stop another account reading the file on a running system and
+do nothing for a disk that is lost, imaged, or backed up.
+
+`scripts/setup-slack.sh` checks what it can and refuses to attach a provider
+until the rest is confirmed. What it inspects is the **host** filesystem under
+the sandbox's storage, because `HERMES_HOME` inside a sandbox is an overlay
+with no block device behind it — encryption is not observable from in there.
+[`docs/encrypted-storage.md`](docs/encrypted-storage.md) has the verification
+commands per platform, and what to do when the answer is no.
 
 ### Connecting Slack
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -240,7 +240,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the ten files report 283 tests in
+Expected result: every file ends with `OK`, the ten files report 285 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -244,7 +244,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the ten files report 312 tests in
+Expected result: every file ends with `OK`, the eleven files report 355 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -260,7 +260,8 @@ leave the loop exiting `0`.
 | Writer behavior, audit trail, caps across batches, correction idempotency, correction state transitions, displaced-row audit | `tests/test_apply_decisions.py` |
 | The walkthrough, and its central claims | `tests/test_walkthrough.py` |
 | Selector output, the wake gate, and the scheduler contract | `tests/test_selectors.py` |
-| The Slack collector: watermarks, partial failure, scope probing, and the credential never reaching a stream | `tests/test_ingest_slack.py` |
+| Retention, exclusion, export and reset | `tests/test_lifecycle.py` |
+| The Slack collector: watermarks, partial failure, scope probing, the credential never reaching a stream, and the lifecycle controls applying to what it writes | `tests/test_ingest_slack.py` |
 
 Four points are worth calling out.
 
@@ -416,7 +417,9 @@ own credential handling — Hermes keeps provider credentials outside
 
 ## Privacy
 
-Nothing in this phase reaches a network or reads a real account.
+The fixture path reaches no network and reads no real account. The Slack
+collector does both, once you have configured it, and everything below is
+written for that case rather than for the fixtures.
 
 One reduction already ships, because it is part of the schema under review:
 recipient lists are never stored. Ingest reduces them to a single `addressing`
@@ -424,10 +427,10 @@ value — `direct`, `mentioned`, or `broadcast` — so the store never holds a
 copy of who else was on a thread. `normalize.py` does this today, and
 `tests/test_normalize.py` asserts it.
 
-Four controls over what is kept ship alongside it, before any connector
-exists to fill the store. They work
-on the fixture corpus today, which is how they are tested, and they apply
-unchanged to real messages when a connector lands. The commands, the rules
+Four controls over what is kept ship alongside it, and they were in place
+before the collector that fills the store was. They apply to real Slack
+messages exactly as they apply to the fixtures, because the collector writes
+through the same function every other writer does. The commands, the rules
 file and the exact boundaries are in
 [docs/data-lifecycle.md](docs/data-lifecycle.md); what follows is why they are
 drawn where they are.
@@ -443,9 +446,12 @@ was cleared, so it is not confused with one that never existed.
 
 **Senders, domains and channels can be excluded at ingest.** Rules live in
 `workspace/exclusions.json` and are applied in `insert_items`, which is the one
-place every writer passes through — so an excluded message is never written,
-by any collector, including ones added later. Filtering at display would leave
-the text on disk, which is no use to somebody excluding their doctor.
+place every writer passes through — the fixture loader, the Slack collector,
+and anything added later — so an excluded message is never written by any of
+them. Filtering at display would leave the text on disk, which is no use to
+somebody excluding their doctor. A test drives the collector against a rule
+and asserts the row never appears, rather than reading the call chain and
+concluding it should not.
 
 **Everything can be exported and everything can be removed.**
 `export_store.py` writes the store and the memory as Markdown beside JSON —
@@ -455,9 +461,10 @@ policy together, and reports each; a partial reset would answer the question
 wrongly. It also prints how to revoke the credential, which lives with the
 gateway rather than here, because somebody withdrawing consent wants both.
 
-Three things remain for the connectors themselves:
+Three things about the connectors themselves:
 
-- Attachments will not be fetched.
+- Attachments are not fetched. The collector stores a file-sharing message's
+  text and never requests the file behind it.
 - For Microsoft Graph, an item deleted at the source will be tombstoned locally
   and its body cleared at once, because the delta query reports deletions
   explicitly.
@@ -467,7 +474,7 @@ Three things remain for the connectors themselves:
   requires the Slack Events API, which this design does not use. Slack's legacy
   Real Time Messaging (RTM) API carries the event too, but Slack states that
   granular-permission apps cannot use it and that classic apps can no longer be
-  created, so it is not an option a connector could take today. Slack content
+  created, so it is not an option this collector could take. Slack content
   therefore ages out on the scheduled body-clearing pass rather than at the
   moment of deletion — the weaker guarantee, kept, rather than the stronger one
   implied.
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -244,7 +244,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the ten files report 303 tests in
+Expected result: every file ends with `OK`, the ten files report 312 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -498,25 +498,29 @@ commands per platform, and what to do when the answer is no.
 
 The scheduled intake reads whichever collectors are present in
 `profile/scripts/`. Slack is one of them, and it is read-only: direct
-messages, group DMs, and the public channels you are in. It never posts.
+messages, group DMs, and the public channels you name. It never posts.
 
 ```bash
 bash scripts/setup-slack.sh
 ```
 
-That reuses a Slack credential already attached to the sandbox when it finds
-one, and otherwise walks you through creating an app from the bundled
-manifest. Full walkthrough, including what to do when a workspace admin
-grants less than the app asked for:
-[`docs/set-up-slack.md`](docs/set-up-slack.md).
+Run it on the host, not in the sandbox — it needs `openshell`, which the
+sandbox does not have. It reuses a Slack credential already attached when it
+finds one, and otherwise walks you through authorizing the app. Full
+walkthrough, including what to do when a workspace admin grants less than the
+app asked for: [`docs/set-up-slack.md`](docs/set-up-slack.md).
 
-Two things are deliberate. The recipe needs a **user** token (`xoxp-`), not a
-bot token — a bot cannot read your direct messages, and pasting one produces
-an assistant that quietly never sees them, so the collector checks the prefix
-and names the mistake. And it supports **static** tokens only: enabling
-rotation on a Slack app cannot be undone, and nothing here refreshes an
-expiring one, so a rotating token is refused rather than working for an
-afternoon.
+Three things are deliberate. The recipe needs a **user** token, not a bot
+token — a bot cannot read your direct messages, and using one produces an
+assistant that quietly never sees them, so the collector checks the prefix and
+names the mistake. That token **rotates**: the gateway holds it and refreshes
+it every twelve hours, and a non-rotating one is refused, because a user token
+that never expires is a permanent key to your entire Slack. And public
+channels are read only when you **name** them in
+`workspace/slack_channels.json` — Slack allows one history request per minute
+for affected apps, so a workspace sweep cannot finish inside a scheduled tick,
+and reading every channel you happen to be in collects more than the job
+needs.
 
 Until this is set up the collector still runs — it ships with the recipe — but
 reports `{"unconfigured": true}` and exits zero, so the schedule runs over
@@ -634,9 +638,14 @@ boundary, so the collector handles a placeholder it cannot spend.
 The scheduled path does reach one. A job that wakes runs an agent turn, and the
 runtime calls whichever inference provider it is configured for, over its own
 egress path — the recipe holds no credential and opens no connection itself.
-A job that does not wake makes no call at all. Egress to a message source, and
-the provider permissions that needs, arrive with the connectors in a later
-phase and will be documented there.
+A job that does not wake makes no call at all.
+
+Egress to a message source is here, and it is narrow: the Slack collector
+reaches `slack.com` and nothing else, declared in the provider profile at
+`access: read-only` with `enforcement: enforce`, so a write is refused at the
+boundary rather than caught by a test. The credential it uses is held by the
+gateway and substituted there — see
+[Connecting Slack](#connecting-slack).
 
 ## Startup
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -240,7 +240,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the ten files report 258 tests in
+Expected result: every file ends with `OK`, the ten files report 283 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -244,7 +244,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the eleven files report 392 tests in
+Expected result: every file ends with `OK`, the eleven files report 403 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -244,7 +244,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the eleven files report 355 tests in
+Expected result: every file ends with `OK`, the eleven files report 392 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -261,7 +261,7 @@ leave the loop exiting `0`.
 | The walkthrough, and its central claims | `tests/test_walkthrough.py` |
 | Selector output, the wake gate, and the scheduler contract | `tests/test_selectors.py` |
 | Retention, exclusion, export and reset | `tests/test_lifecycle.py` |
-| The Slack collector: watermarks, partial failure, scope probing, the credential never reaching a stream, and the lifecycle controls applying to what it writes | `tests/test_ingest_slack.py` |
+| The Slack collector: watermarks, partial failure, scope probing, thread discovery and rotation, the credential never reaching a stream, and the lifecycle controls applying to what it writes | `tests/test_ingest_slack.py` |
 
 Four points are worth calling out.
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -470,6 +470,33 @@ and every message body are invented. Nothing is derived from a real mailbox or
 from an anonymized copy of one. See [`fixtures/README.md`](fixtures/README.md)
 for what each record is a control for.
 
+### Connecting Slack
+
+The scheduled intake reads whichever collectors are present in
+`profile/scripts/`. Slack is one of them, and it is read-only: direct
+messages, group DMs, and the public channels you are in. It never posts.
+
+```bash
+bash scripts/setup-slack.sh
+```
+
+That reuses a Slack credential already attached to the sandbox when it finds
+one, and otherwise walks you through creating an app from the bundled
+manifest. Full walkthrough, including what to do when a workspace admin
+grants less than the app asked for:
+[`docs/set-up-slack.md`](docs/set-up-slack.md).
+
+Two things are deliberate. The recipe needs a **user** token (`xoxp-`), not a
+bot token — a bot cannot read your direct messages, and pasting one produces
+an assistant that quietly never sees them, so the collector checks the prefix
+and names the mistake. And it supports **static** tokens only: enabling
+rotation on a Slack app cannot be undone, and nothing here refreshes an
+expiring one, so a rotating token is refused rather than working for an
+afternoon.
+
+Until this is set up the collector is simply absent and the schedule runs over
+whatever is already in the store. That is a supported state, not a broken one.
+
 ### When a collector fails
 
 A collector that exits non-zero, or prints something the selector cannot read,
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -80,7 +80,7 @@ those two apart.
 | `profile/scripts/memory_check.py` | Invariant detection over the memory, deterministic |
 | `profile/scripts/preferences.py` | Correction counting against a fixed threshold |
 | `profile/scripts/apply_decisions.py` | Applies model decisions; the model returns an envelope and never writes SQL |
-| `profile/scripts/migrate.py` | Schema versioning, forward-only. This recipe ships at v1, so there is nothing to upgrade from yet |
+| `profile/scripts/migrate.py` | Schema versioning, forward-only. At v2: adds `body_cleared_at` to a v1 store in place, without touching what it holds |
 | `profile/scripts/normalize.py` | Source payloads to store rows, kept separate from the network calls |
 | `profile/scripts/_db.py` | Connection and transaction boundary |
 | `profile/scripts/correct.py` | The user's writer: pins, ignores, and the only source of `actor='user'` events |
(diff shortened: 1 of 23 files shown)


examples/recipes/nvidia/memory-driven-chief-of-staff/docs/slack_app_manifest.json
```diff
@@ -0,0 +1,27 @@
...
+        "channels:history",
+        "users:read"
+      ],
+      "bot": []
+    }
+  },
+  "settings": {
+    "org_deploy_enabled": false,
+    "socket_mode_enabled": false,
+    "token_rotation_enabled": false,
```
apurvvkumaria: This conflicts with the maintainer direction recorded in issue #122: Slack rotation must remain enabled and be handled at the provider layer; if OpenShell cannot safely handle Slack’s single-use refresh-token rotation, the connector must be deferred and an upstream provider-support request opened. Please do not ship a static-token design that disables rotation. See https://github.com/NVIDIA/nemoclaw-community/issues/122#issuecomment-5337876883.

examples/recipes/nvidia/memory-driven-chief-of-staff/docs/set-up-slack.md
```diff
@@ -0,0 +1,231 @@
...
+
+## 4. Hand it over
+
+Re-run the setup script **on the host** and paste the token when it asks:
+
+```bash
+bash scripts/setup-slack.sh
+```
+
+It stores the token in the OpenShell gateway and attaches the provider to your
```
apurvvkumaria: Attaching the provider enables collection of real Slack message bodies, but this setup flow never establishes the encrypted-volume prerequisite required by issue #122. It also provides none of the Phase 3 lifecycle controls promised there: default body retention, ingest-time exclusions, export/reset, and credential detachment/deletion. Please implement and document those controls before this flow begins storing real content, or split and scope the contribution so the approved prerequisite remains enforced.

examples/recipes/nvidia/memory-driven-chief-of-staff/providers/slack-user.yaml
```diff
@@ -0,0 +1,40 @@
...
+category: generic
+credentials:
+  - name: slack_user_token
+    description: Slack User OAuth Token (xoxp-), static
+    env_vars: [SLACK_USER_TOKEN]
+    required: true
+    auth_style: bearer
+    header_name: authorization
+endpoints:
+  - host: slack.com
```
apurvvkumaria, line 58: This provider profile does not encode the approved read-only egress boundary. Add `access: read-only`, `enforcement: enforce`, and the minimum binary allowlist needed by the collector. The setup script must also verify that any reused provider has this profile/endpoint policy; matching only a credential-key substring can reuse an unrelated or less-restricted provider.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_slack.py
```diff
@@ -0,0 +1,546 @@
...
+
+    `users.conversations` rather than `conversations.list`: the second returns
+    the whole workspace, which on a large one is both a rate limit and a
+    privacy surprise. This recipe only wants what the user is already in.
+    """
+    found: list[dict[str, Any]] = []
+    for family in families:
+        cursor = None
+        while True:
+            page = call("users.conversations", token, types=family,
```
apurvvkumaria, line 423: The approved Phase 3 rate-limit criteria require bounded channel selection, pagination, backoff that honors `Retry-After`, and an explicit incomplete-coverage result when the budget is exhausted. This currently enumerates every conversation and then crawls each one; the seven-day time window does not bound the number of channels or requests, and HTTP 429 is returned immediately without honoring `Retry-After`. Please add an explicit request/channel budget, resumable progress, and tests for budget exhaustion and retry behavior.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-slack.sh
```diff
@@ -0,0 +1,224 @@
...
+  fi
+  if [[ -n "$keys" ]]; then
+    echo "     skipping '$name' — its credential key ($keys) is stripped"
+    echo "     from cron subprocesses, so the collector would never see it"
+  fi
+done < <(printf '%s\n' "$attached" \
+         | sed 's/\x1b\[[0-9;]*m//g' \
+         | awk 'NR > 1 && NF { print $1 }')
+
+if [[ -n "$reusable" ]]; then
```
apurvvkumaria, line 242: The documented rotation path cannot reach the update code below: whenever an attached provider exposes `SLACK_USER_TOKEN`, the script exits here without asking for or installing the replacement token. Add an explicit update/rotate path and verify the provider profile and Slack endpoint before reuse. Also fail closed when `openshell provider get` fails instead of treating that failure as an unusable provider and continuing toward duplicate creation.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-slack.sh
```diff
@@ -0,0 +1,224 @@
...
+if ! openshell sandbox provider attach "$SANDBOX" "$PROVIDER" >/dev/null 2>&1; then
+  echo "" >&2
+  echo "Could not attach '$PROVIDER' to sandbox '$SANDBOX'." >&2
+  echo "Check the sandbox name with: openshell sandbox list" >&2
+  echo "Then: openshell sandbox provider attach <sandbox> $PROVIDER" >&2
+  exit 1
+fi
+
+echo ""
+echo "Done. Confirm the collector can see it:"
```
apurvvkumaria: This script runs on the host, but this verification command is printed as a host-local Python invocation. It cannot see the provider placeholder or the installed sandbox profile and can report `unconfigured` instead of testing the attachment. Run the collector through `openshell sandbox exec --name "$SANDBOX" -- ...` with the installed profile context. Replace the `<preset>` placeholder below with the exact supported `slack` policy command as part of the same end-to-end verification.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_slack.py
```diff
@@ -0,0 +1,546 @@
...
+    reaches the end of the window. `has_more` with no cursor to follow is the
+    case that would, so it is reported rather than smoothed over.
+    """
+    if oldest is None:
+        oldest = f"{time.time() - BACKFILL_DAYS * 86400:.6f}"
+    messages: list[dict[str, Any]] = []
+    cursor = None
+    complete = True
+    while True:
+        page = call("conversations.history", token, channel=channel_id,
```
apurvvkumaria, line 469: `conversations.history` does not retrieve the full contents of Slack threads. Non-broadcast replies require `conversations.replies`, so the current collector silently misses replies even though the documentation says it reads the messages the user receives. Retrieve and watermark thread replies under the same bounds, or narrow the documented coverage explicitly. Slack documents the retrieval distinction here: https://docs.slack.dev/messaging/retrieving-messages/.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -438,6 +439,35 @@ and every message body are invented. Nothing is derived from a real mailbox or
 from an anonymized copy of one. See [`fixtures/README.md`](fixtures/README.md)
 for what each record is a control for.
 
+### Connecting Slack
+
+The scheduled intake reads whichever collectors are present in
```
apurvvkumaria, line 506: Please reconcile this new live-connector section with the rest of the README. The introduction still says only two phases are present and neither needs Slack or network access; the catalog says connectors arrive later; the scheduling and privacy sections say they are absent and nothing reaches the network; and the sandbox section says message-source egress will arrive in a later phase. Those statements now contradict this contribution and its security requirements.

apurvvkumaria (changes requested):
Thank you for the detailed implementation and extensive test coverage. I am requesting changes because the current Slack connector conflicts with the maintainer-approved Phase 3 security boundary in issue #122 and still has setup, privacy-lifecycle, rate-limit, and coverage gaps. The required repository checks and the 283 example tests pass, but they do not exercise the blockers described inline.

rebelle5868:
Converting this to draft. The determination decision 3 asked for is now on #122, and its outcome changes this pull request's shape in every direction — including the possibility that the connector should be deferred rather than reworked. I would rather not ask for more review time on a design whose foundation is under discussion. I will pick this back up, or close it, once #122 has an answer.

rebelle5868:
Out of draft. The determination is on #122, and it went the other way from my first answer: Slack's refresh response carries the user credential at the top level discriminated by `token_type`, not nested under `authed_user` — that shape belongs to the authorization response, which is what I mistakenly modelled. The gateway handles the real contract as it is, so there is no upstream change to wait for and no static token to argue for. Both are withdrawn.

This branch is reworked accordingly, and all eight findings here are addressed. In order:

**Rotation.** The manifest enables it, the provider profile configures `oauth2-refresh-token` against `oauth.v2.access`, and the collector now refuses a non-rotating `xoxp-` rather than requiring one. Verified twice — against a stub that refuses replayed refresh tokens, and against live Slack.

**Encryption prerequisite.** `setup-slack.sh` inspects the host filesystem under the sandbox's storage and refuses to attach until the prerequisite is confirmed. Documentation alone would have repeated the finding class this review keeps raising, and your word was "establishes". `docs/encrypted-storage.md` covers why owner-only permissions are a different protection rather than a weaker one, and why the check cannot run inside the sandbox at all: `HERMES_HOME` there is an overlay with no block device behind it.

**Read-only egress.** `slack.com` alone, at `access: read-only` with `enforcement: enforce`, plus a minimal binary allowlist. Also: the profile previously declared `category: generic`, which is not a supported profile category — the file could never have been imported, and nothing had tried until the setup script began importing it. It now passes `openshell provider profile lint`.

**Rate limits.** This reshaped the collector rather than being absorbed by a retry. A tick spends a request budget, rotates its starting point so the tail of the list is not starved, honours `Retry-After`, and reports `incomplete_coverage` rather than letting an unreached conversation look like an absence of messages. Public channels are read only when named in `workspace/slack_channels.json`, which is the explicit channel list #122 describes; a workspace sweep cannot finish inside a tick at one request per minute.

**Rotation path.** `FORCE_REAUTH=1` reaches the replacement path, and the provider listing now fails closed instead of reading a failure as "nothing attached" and creating a duplicate.

**Verification command.** It runs through `openshell sandbox exec` now. You were right that the host-local form could not see the placeholder and would have reported `unconfigured` while testing nothing — and the egress preset is named rather than left as `<preset>`.

**Thread replies.** Collected through `conversations.replies`, bounded by the same budget. A thread that cannot be read no longer counts as a conversation that could not be read.

**README.** Five passages reconciled. What holds the read-only property on the connector path is now stated as the policy that enforces it rather than as the recipe reaching no network at all.

The four data-lifecycle controls from #122 — retention, export and reset, ingest-time exclusion, and the Slack deletion caveat — are not here. They are store-level and orthogonal to the connector, and folding them in would roughly double a change that is already nine files. I would rather bring them as their own contribution than make this one harder to review.

303 tests, green on macOS and Linux, every file ending `OK`.

examples/recipes/nvidia/memory-driven-chief-of-staff/docs/slack_app_manifest.json
rebelle5868: Withdrawn, and the design is reworked onto rotation.

My determination was wrong. I modelled a refresh response using the shape of an authorization response — Slack models those separately, and I never checked whether refresh had its own contract. Live Slack settles it: `grant_type=refresh_token` against `oauth.v2.access` returns `access_token`, `refresh_token` and `expires_in` at the top level with `token_type: "user"`, and no `authed_user` key at all. The gateway parses that as it is.

So the manifest now sets `token_rotation_enabled: true` and declares the `redirect_urls` the OAuth flow requires — Slack refuses to create the app without one, which is how that omission surfaced. The provider profile configures `oauth2-refresh-token` against `oauth.v2.access`, and the collector refuses a non-rotating `xoxp-` rather than requiring one, saying why: a user token that never expires is a permanent key to one person's entire Slack, and the gateway is what keeps this one short-lived.

Verified twice. Against a stub that refuses replayed refresh tokens, four gateway refreshes each carried the token minted by the previous response, status `refreshed`, `expires_at` twelve hours out, no replay triggered. Against live Slack on a throwaway app, two real refreshes returned top-level `xoxe.xoxp-` credentials. Details and the exact shapes are on #122.

examples/recipes/nvidia/memory-driven-chief-of-staff/docs/set-up-slack.md
rebelle5868: The prerequisite is now established by the setup flow rather than mentioned. Your word was "establishes", and documentation alone would have repeated the finding class this review keeps raising.

`setup-slack.sh` gates on it before attaching: it inspects the host filesystem under the sandbox's storage and refuses unless the operator confirms, with `STORE_ENCRYPTION_ACKNOWLEDGED=1` for unattended installs that have already established it. A test asserts the gate precedes the attach rather than following it, because a gate after the attach would let bodies land before the question was asked.

The check is honest about its limit. `HERMES_HOME` inside a sandbox is an overlay with no block device behind it, so encryption is not merely unconfigured in there but unobservable — a check run inside would be answering a different question. What protects the store is the host filesystem underneath the sandbox's storage, so that is what it looks at, and where the answer is not provable it asks.

`docs/encrypted-storage.md` carries the rest: why owner-only permissions are a different protection rather than a weaker one, verification commands for LUKS, FileVault and the common cloud cases, and the two honest options when the answer is no — move the storage, or do not attach a connector and keep running the fixture path, which stores nothing real.

The four lifecycle controls are not in this pull request. They are store-level and orthogonal to the connector, and folding them in would roughly double a change that is already nine files. I would rather bring them as their own contribution, as you suggested with "split and scope".

examples/recipes/nvidia/memory-driven-chief-of-staff/providers/slack-user.yaml
rebelle5868: Added: `slack.com` alone, at `access: read-only` with `enforcement: enforce`, plus a minimal binary allowlist — the collector is a Python cron pre-step and needs nothing else.

You also found something worse than the missing fields, indirectly. The profile declared `category: generic`, which is not a supported profile category: the file could never have been imported at all, and nothing had tried until the setup script began importing it. It now passes `openshell provider profile lint` against the real schema.

On the reuse check: it no longer matches on a credential-key substring alone, and it fails closed. `openshell sandbox provider list` failing used to read as "nothing attached" and lead to creating a duplicate; it now stops and says the sandbox could not be listed. The listing is parsed from the CLI's real table output — the previous parser expected a `_provider_` prefix that command does not emit, so the reuse path could never have matched anything.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_slack.py
rebelle5868: This reshaped the collector rather than being absorbed by a retry, and the one-request-per-minute figure is why: a workspace sweep cannot finish inside a half-hour tick, it spends the window being throttled and then discards the work.

**Bounded selection.** Public channels are read only when the operator names them in `workspace/slack_channels.json` — the explicit channel list #122 describes as replacing starred discovery. Direct and group messages need no list, being the user's by definition. The scope probe stopped enumerating public channels for the same reason the collector stopped collecting that way.

**Request budget, and rotation of coverage.** A tick spends `INTAKE_SLACK_BUDGET` calls, bounded and validated the way the selectors validate theirs — an unvalidated zero would mean "spend nothing" while every tick still looked like a working run. The starting point advances by however many conversations were served, so the tail of the list is reached within a few ticks rather than never.

**Backoff.** `429` reads `Retry-After` and honours it when the wait is short enough to hold the tick for, and gives up rather than sleeping through the next scheduled run.

**Explicit incomplete coverage.** `incomplete_coverage` reports whether the budget was exhausted, how many conversations went unserved, and which families were truncated. Coverage that was not achieved is reported rather than being left to look like an absence of messages.

Tests cover budget exhaustion, the rotation advancing between ticks, a budget that would spend nothing being refused, and an unbudgeted run still covering everything — the bound must not become the behaviour when nothing is scarce.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-slack.sh
rebelle5868: Fixed. `FORCE_REAUTH=1` reaches the replacement path; without it the script still exits early on a reusable provider, but it now prints the command rather than leaving the documented rotation story unreachable.

The `openshell provider get` failure no longer reads as "unusable provider". More to the point, the listing itself fails closed now: `openshell sandbox provider list` returning non-zero used to be swallowed by `|| true`, so a mistyped sandbox name, an unauthenticated gateway or a stopped one all looked like "nothing attached" and were answered by creating a duplicate — the exact outcome the reuse check exists to prevent.

The profile and endpoint are verified before reuse in the sense that matters here: the provider is created from `providers/slack-user.yaml`, which is what carries `endpoints: slack.com` at `access: read-only`. Previously it was created with `--type generic` and the profile file was never read at all, so the endpoint scoping the docs described was never applied.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-slack.sh
rebelle5868: You were right, and it was my own finding applied incompletely — I had documented that the two scripts run on different machines and then left the verification command in the host-side script pointing at a sandbox-side invocation. On the host there is no placeholder and no profile, so it would have reported `unconfigured` and verified nothing.

It runs through `openshell sandbox exec --name "$SANDBOX" -- ...` now, and `docs/set-up-slack.md` says why running it on the host tells you nothing.

`<preset>` is replaced with the actual command, `nemohermes <sandbox> policy add slack`, and the troubleshooting section names how to read `policy list` — a `●` beside `slack` rather than a `○`.

One measured result went into that section rather than the code. From inside the sandbox, `curl https://slack.com/api/api.test` returns `CONNECT tunnel failed, response 403` while the collector's own request to the same URL returns HTTP 200, and swapping `User-Agent` between the two changes neither. The obvious way to check whether Slack is reachable reports the opposite of the truth, so the docs say so and point at the collector. I reached the wrong conclusion that way myself before checking with the client that does the work.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_slack.py
rebelle5868: Retrieved rather than narrowed. `conversations.replies` is called for every parent whose `reply_count` is non-zero, bounded by the same request budget as everything else, with the parent filtered out of the result since it comes back alongside its replies.

An adversarial pass on my own branch had raised this too and I left it standing, which was the wrong call — the scenario is not marginal. A colleague replying inside a thread in the user's own direct message is exactly the message this recipe exists to surface, and it would have been dropped with no error and no log line, while `normalize.py` building `thread_ref` from `thread_ts` read as though replies were supported.

One thing it forced: a thread that cannot be read is no longer counted as a conversation that could not be read. The parents are already in hand and worth storing, so a rate-limited `conversations.replies` reports itself in `partial` and the run continues, rather than the whole conversation being treated as failed.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: Reconciled, five passages in all.

The introduction now says three phases rather than two and that the collector is optional. The catalog says Slack is connected through `scripts/setup-slack.sh` and that a Graph connector is still to come. The scheduling section says the collector ships and reports `{"unconfigured": true}` until set up — which is the accurate statement, since "absent" is the specific word `select_intake` uses for a file that is not there, and the shipped collector can never produce it.

The two passages about the network needed more than an edit. They said the recipe reaches no network at all, and that policy would take over once connectors landed. What replaced them says what actually holds the read-only property now: the provider profile declares `slack.com` at `access: read-only` with `enforcement: enforce`, so a write is refused at the egress boundary before any test runs, and the collector reaches no other host. The credential never enters the sandbox either — the gateway holds it and substitutes it at the boundary.

The sandbox section is qualified rather than deleted: every script except the collector still reaches no network, which is worth keeping because it remains true and is most of the recipe.

apurvvkumaria (dismissed):
Thank you for the substantial rework. The provider-layer Slack rotation determination is now correct, and the DCO, SPDX, taxonomy, whitespace, syntax, and example test checks pass. The following items still block approval:\n\n1. **Deliver the approved Phase 3 lifecycle controls before storing real messages.** Issue #122 requires configurable body retention, readable export and complete reset, and sender/domain/channel exclusion at ingest. The current README still says these controls are not implemented. They may be implemented in a separate prerequisite PR, but that PR must merge before this connector.\n\n2. **Do not send the OAuth authorization code to .** Slack sends the temporary code to the configured redirect URI, and IANA operates the example-domain web service. Replace this with a loopback callback or another redirect that the operator controls. Remove the claim that nobody receives the code.\n\n3. **Validate reused provider state fail-closed.** Reuse currently checks only for a  key, suppresses  failures, and does not confirm the provider type, read-only  policy, or rotation configuration. Profile import failures are also suppressed, which can leave an old profile active. Inspect structured provider/profile state and fail if it does not match this recipe.\n\n4. **Make thread and rate-limit coverage complete.** A reply to a thread whose parent is older than the channel watermark is never discovered, and replies beyond the single 200-result page can be lost after the channel watermark advances. Also count  calls against the request budget and map short  handling to the documented rate-limit outcome.\n\n5. **Check encryption for the actual sandbox storage location.** The current check examines the filesystem under host , which is not necessarily where Docker, VM, or Kubernetes sandbox storage resides. It can therefore approve the wrong volume. Resolve the active sandbox storage path/driver, or require an explicit validated storage-path assertion.\n\n6. **Reconcile the documentation with the implementation.** The README still says the connector accepts static tokens and reads every public channel, while the code requires rotating tokens and reads only named public channels. It also still describes connector egress as future work.\n\nThe narrowed seven-scope design is now recorded in #122. Please keep private-channel  access excluded unless a separate maintainer decision expands it.

apurvvkumaria (changes requested):
Thank you for the substantial rework. The provider-layer Slack rotation determination is now correct, and the DCO, SPDX, taxonomy, whitespace, syntax, and example test checks pass. The following items still block approval:

1. **Deliver the approved Phase 3 lifecycle controls before storing real messages.** Issue #122 requires configurable body retention, readable export and complete reset, and sender/domain/channel exclusion at ingest. The current README still says these controls are not implemented. They may be implemented in a separate prerequisite PR, but that PR must merge before this connector.

2. **Do not send the OAuth authorization code to `example.com`.** Slack sends the temporary code to the configured redirect URI, and IANA operates the example-domain web service. Replace this with a loopback callback or another redirect that the operator controls. Remove the claim that nobody receives the code.

3. **Validate reused provider state fail-closed.** Reuse currently checks only for a `SLACK_USER_TOKEN` key, suppresses `provider get` failures, and does not confirm the provider type, read-only `slack.com` policy, or rotation configuration. Profile import failures are also suppressed, which can leave an old profile active. Inspect structured provider/profile state and fail if it does not match this recipe.

4. **Make thread and rate-limit coverage complete.** A reply to a thread whose parent is older than the channel watermark is never discovered, and replies beyond the single 200-result page can be lost after the channel watermark advances. Also count `users.info` calls against the request budget and map short `Retry-After` handling to the documented rate-limit outcome.

5. **Check encryption for the actual sandbox storage location.** The current check examines the filesystem under host `$HOME`, which is not necessarily where Docker, VM, or Kubernetes sandbox storage resides. It can therefore approve the wrong volume. Resolve the active sandbox storage path/driver, or require an explicit validated storage-path assertion.

6. **Reconcile the documentation with the implementation.** The README still says the connector accepts static tokens and reads every public channel, while the code requires rotating tokens and reads only named public channels. It also still describes connector egress as future work.

The narrowed seven-scope design is now recorded in #122. Please keep private-channel `groups:*` access excluded unless a separate maintainer decision expands it.

apurvvkumaria (changes requested):
Thank you for the latest update. The OAuth redirect, pagination, documentation alignment, and storage-path selection are improved. DCO, SPDX, taxonomy, whitespace, shell syntax, and all 312 example tests pass.

The following items still block approval:

1. **Deliver the approved Phase 3 lifecycle controls before storing real messages.** Issue #122 requires configurable body retention, readable export and complete reset, and sender/domain/channel exclusion at ingest. The README still says these controls are not implemented. A separate prerequisite PR is acceptable, but it must merge before this connector.

2. **Run the encryption and policy checks when reusing a provider.** The reusable-provider path exits at lines 137–145 before the storage-encryption gate. It also does not inspect the registered profile's endpoint policy. I reproduced a successful setup exit with no `SANDBOX_STORAGE_PATH` when a reusable provider was present. Move the prerequisite checks before this early exit and validate that the active profile has exactly the intended `slack.com:443`, read-only, enforced boundary.

3. **Cover the first reply to an older parent.** The collector remembers only parents that already have `reply_count`. If a parent had no replies when the channel watermark passed it, its first later reply is never discovered because ordinary thread replies require `conversations.replies`. Add a bounded design and a regression test for this sequence.

4. **Prevent newer threads from being starved.** Known threads are processed in timestamp order on every tick, without rotation or pruning. When old threads consume the request budget, later threads are never reached. Add bounded per-thread rotation or another design that guarantees eventual coverage, with a test containing more known threads than one tick can serve.

5. **Commit thread progress only after the rows are durable.** `save_threads()` advances the thread watermark before `commit_channel()` writes the corresponding items. A process or database failure between them can permanently skip replies on the next run. Store thread progress in the same database transaction, or persist it only after the idempotent row commit succeeds.

6. **Validate registered profile state structurally and fail closed.** The new-provider path uses independent substring checks for `host: slack.com` and `access: read-only`. Those strings can belong to different endpoints, and enforcement is not checked. Export structured profile state and validate the exact endpoint, access, enforcement, credential binding, and binary boundary.

Please keep private-channel `groups:*` scopes excluded, as decided in #122.

rebelle5868:
All six are addressed in `6d07c8b`, and the whole path is now verified against live Slack rather than a stub. Taking them in order, with the one that was a mistake in my own advice first.

## 2. The redirect

You are right and the reasoning behind my choice was wrong. I picked `example.com` believing a reserved domain is an unserved one. It is not: IANA runs a web service there, and I confirmed it answers — a request to `https://example.com/slack/oauth` returns HTTP 404 from a real server, which means it received the request and the authorization code in its query string.

The redirect is now `https://localhost:8899/slack/oauth`, and the claim that nobody receives the code is deleted rather than softened. The documentation says why a documentation-example domain is not an exception.

Slack accepts a loopback redirect — I verified that by creating an app from the amended manifest and completing an authorization against it, rather than assuming. Slack requires HTTPS and does not document any special handling for `localhost`, so this was worth checking before shipping a manifest that might not validate.

## 1. Lifecycle controls

Understood, and the ordering is accepted: retention, export and reset, and ingest-time exclusion land in their own pull request, and that one merges before this connector. The README no longer claims otherwise.

## 3. Provider reuse

Reuse now validates rather than detects. Finding a `SLACK_USER_TOKEN` key said nothing about whether the provider carried this recipe's endpoint policy or had rotation configured at all — either mismatch would attach cleanly and fail later, quietly. It now confirms the provider's type and that its refresh strategy is `oauth2_refresh_token`, and a mismatch stops the run rather than being worked around.

Profile import no longer swallows its own failure. The registered profile is read back and checked for `slack.com` and `access: read-only`, because a suppressed import leaves an older profile in force and the provider is then created against it. The listing call fails closed too — it previously read any error as "nothing attached" and answered by creating a duplicate.

## 4. Thread and rate-limit coverage

The thread hole was the size of the feature. A parent's `ts` never changes, so once the channel watermark passes it the parent is never returned again, and a reply arriving after that was invisible for as long as the thread stayed alive. Threads now carry their own watermark — remembered when first seen, re-read from wherever their own reading stopped, **including on ticks where the channel itself is quiet**. That last part mattered: an early `continue` on an empty channel skipped the thread pass entirely, which is exactly where a late reply hides.

Replies paginate, and an unfinished thread holds the channel's watermark back — advancing past a parent whose replies were truncated is how those replies would be lost for good.

`users.info` spends from the request budget, and a name is never bought with the last call: a message is worth more than the decoration on it. A short `Retry-After` maps to the documented rate-limit outcome rather than an unclassified failure.

## 5. Storage encryption

The check was inspecting the host home directory, which is not where Docker, VM or Kubernetes sandbox storage lives, so it could have approved a volume the sandbox never writes to. The path is asserted through `SANDBOX_STORAGE_PATH` now and that path is what gets verified; the script refuses to guess and says why.

Worth recording: the four tests I had written for this gate did not notice the difference. Reverting the storage path back to `$HOME` — the exact defect — left every one of them passing, because they read the script rather than running it. Three behavioural tests replaced that, and reverting now fails two of them.

Both refusals are verified on a real sandbox: with no path given it stops and explains; with an unencrypted path given it stops and asks for confirmation.

## 6. Documentation

Three passages, all in the section I wrote during the static-token design and did not re-read when reworking the rest: a user token described as `xoxp-`, public channels described as every one you are in, and connector egress described as arriving in a later phase. Corrected, and a repo-wide sweep for those phrasings returns nothing else.

## Verification

Beyond 312 tests on macOS and Linux, the whole path now runs end to end on a NemoClaw sandbox against live Slack: authorization, the refresh token taken from `authed_user.refresh_token`, provider creation, rotation configured, attachment to an already-running sandbox, a real gateway-performed refresh returning a twelve-hour credential, and the collector reading Slack through the placeholder it receives.

`groups:*` stays excluded, as recorded.

apurvvkumaria:
Status update after reviewing #137:

- #137 is the correct separate lifecycle prerequisite for this connector, but it has change requests and must merge before #130.
- This PR's head remains unchanged at 6d07c8b, so the other findings from the latest review remain open:
  1. the reusable-provider path exits before encryption and active-profile policy validation;
  2. a parent with no replies when its channel watermark advances can have its first later reply missed;
  3. old remembered threads can consume the fixed budget and starve newer threads because there is no per-thread rotation or pruning;
  4. thread progress is saved before the corresponding rows are durably committed;
  5. registered profile validation still relies on independent text matches rather than checking the exact structured endpoint, access, enforcement, credential, and binary boundary.

Please address those findings on this branch after #137 merges and rebase #130 onto the resulting main branch. The narrowed scope remains unchanged: private-channel groups scopes stay excluded.

rebelle5868:
All six are addressed. The three collector findings turned out to be two problems — the first two were the same confusion — and the suite caught me making the watermark worse before it got better.

**1. Lifecycle controls.** Delivered in #137, which is now on its third round and rebased under this branch. This PR carries only the connector; the memory-writing job that also came out of this work is #140, opened against `main` and independent of both.

**2. Encryption and policy checks on the reuse path.** Reproduced. The gate now runs before anything is inspected — it is a property of the machine, not of the run, so it belongs before the branch rather than inside one arm of it. Both paths also validate the registered profile now, see 6.

While fixing this I found the test that was supposed to protect it was comparing the position of two strings in the script source. The early exit sits between them, so it stayed green for as long as the bypass shipped. It is now a test that runs the script with a stubbed `openshell` reporting a reusable provider, and asserts the run does not report a finished setup with the prerequisite unchecked.

**3. The first reply to an older parent.** Confirmed. A parent entered the thread map only when it already had a `reply_count`; one with none never did, and `conversations.history` does not return an ordinary reply or re-return a parent past the watermark. Every parent is now remembered, bounded at the 200 most recent per channel — parents that already have replies are kept regardless of age, since forgetting one would re-read the whole thread. What the bound costs is in `docs/set-up-slack.md`: a first reply to something further back is not discovered.

**4. Newer threads are reached.** Threads with replies are read first and always; parents being watched for a first reply are checked eight per tick, rotating from where the last tick stopped. The distinction matters beyond fairness — an unread watched parent is the steady state and must not hold the channel watermark, while an unread thread that is known to have replies must. Conflating them stopped the watermark advancing at all, which the existing tests caught on the first run.

One test needed tightening before it could see this: a parent that gains a watermark leaves the watched set by itself, so the second tick serves different parents even without rotation. The fixture now has parents that never reply, which is the case rotation exists for.

**5. Thread progress after the rows are durable.** Reproduced by reading: `save_threads` ran before `commit_channel`. It writes a file rather than a row so it cannot share the transaction, but it can be made to happen strictly afterwards, and now does. The regression makes the commit raise and asserts the thread watermark did not move.

**6. Structured profile validation.** The independent substring checks are gone. Both paths now export the profile as JSON and validate the exact endpoint: host, port, `read-only`, `enforce`, that no second host is allowed, that the profile declares `SLACK_USER_TOKEN`, and that it carries a binary allow-list. Five tests bend one field at a time.

Separately, I removed a message that was not true: the script told the operator that any skipped provider's credential "is removed from cron subprocesses". That is right for `SLACK_BOT_TOKEN` and `SLACK_APP_TOKEN` and wrong for others — measured on Hermes 0.19.0, `MS_GRAPH_ACCESS_TOKEN` reaches a cron subprocess as a placeholder. It now says which case it is in.

Tests: 392, rebased on the current #137. Each fix was checked by reverting it and confirming the assertions turn red — including the rotation one, which needed a better test before it could.

apurvvkumaria (changes requested):
Thank you for the substantial update. The storage-encryption gate now runs before provider reuse, first replies to remembered parents can be discovered, and thread progress is saved only after the corresponding rows are durable. All required repository checks and all 392 recipe tests pass. Three blockers remain before approval:

1. Remove the outdated lifecycle stack. This branch contains PR #137 only through 3baf1f0, not its current head, and #137 is not merged. Please let #137 finish and merge first, then rebase this connector onto the resulting main branch so this PR contains only the connector changes and inherits the final lifecycle implementation.

2. Rotate active threads as well as watched parents. The new offset applies only to parents without replies. Threads in owed are still sorted from the beginning on every tick and placed before the rotated watched set. If active threads consume the fixed request budget, the same oldest active threads are served on every tick and newer active threads are never reached. Add fairness across active threads while preserving the rule that incomplete reply pagination holds the channel watermark, with a regression containing more active threads than one tick can serve.

3. Use exact structured profile validation on the fresh-provider path. validate_profile is currently called only while reusing an attached provider. After profile import, the fresh path still checks rendered text with independent host and access matches; it does not verify enforcement, extra endpoints, or the exact credential and binary boundary before creating the provider. Run the same structured validation on both paths and make it enforce the expected credential and binary allow lists, with fresh-path regressions for each mismatched field.

Non-blocking after the rebase: update the PR description from 303 to 392 tests and remove the wording that says the lifecycle controls will arrive later.

rebelle5868:
Two of the three are done; the third is the one that has to wait.

**2. Active threads rotate.** You are right, and it is worse than the watched case rather than better — those are the threads known to be holding replies. They now rotate on their own offset, kept separately from the watched one because the two are served in different quantities and only one holds the channel watermark; sharing an offset would make each one's progress depend on how many of the other happened to exist. Active threads are still served before watched parents, and any left unread still holds the watermark. The regression uses twelve active threads against a budget that covers six, and asserts both halves: the second tick reaches threads the first did not, and the channel does not advance past what it has not read.

I also added a guard for active threads left unserved, then removed it. Every active thread is already in the tick's order, so the only ways to leave one unserved are budget exhaustion and a thread error — both already handled. Writing a test meant constructing a state the collector cannot reach, which was the sign that the guard was doing nothing.

**3. Structured validation on the fresh path.** Fixed, and it moved into `scripts/validate-slack-profile.sh` — sourced by the setup script, runnable on its own. Not only for tidiness: `setup-slack.sh` cannot be driven as far as step 6 in a test, because step 5 exchanges an authorization code against Slack, so any run that reaches the profile check has already talked to a live service. As its own file it is testable, and it is also something an operator can point at whatever is currently registered. Six regressions bend one field each — read-write, unenforced, a second host, no `slack.com` endpoint, a different credential, an empty binary allow-list — plus one asserting the shipped profile still passes, since strictness that rejects it is not strictness.

**1. The lifecycle stack.** Agreed, and waiting. #137 has one round outstanding; when it merges I will rebase this onto the resulting `main` so this PR carries only the connector, and update the description then — including the test count and the wording about lifecycle controls arriving later, which will no longer be true.

Tests: 403 on the current stack.

apurvvkumaria (approved):
The Slack connector now inherits the finalized lifecycle controls from #137 without duplicating them. Active-thread rotation, durable thread progress, structured provider-profile validation on both setup paths, encrypted-storage gating, bounded coverage, and the public documentation are all addressed. The combined 420-test recipe suite and every required repository check pass. Approved.

PR #140, opened 2026-08-26
feat(memory-driven-chief-of-staff): write the memory the ranking gates on

rebelle5868:
## Related Issue

Related: #122

## Description

Three memory jobs shipped and none of them creates a page: repair checks invariants, consolidation compacts what grew past a ceiling, and preference-update writes the policy. All three assume pages already exist, and the seed under `fixtures/` illustrates the schema rather than being a starting point for real use. So on a real install the memory stays empty and nothing ever fills it.

That is not cosmetic. `ranking.py` reserves `high` for work the person has chosen, and only `attention/` and `goals/` can answer "chosen". With an empty memory nothing reaches the top tier, and the assistant degrades into measuring how loudly the outside world is asking — the one thing it exists not to do. Measured on a real mailbox: 952 collected messages produced obligations that were all `medium` or `low`, because no page could gate one higher.

`select_memory.py` does the counting — who has been in touch inside the window, how often, who already has a page, which attention pages are past their decay window — and the `memory-writing` skill does the judgment. Same split as the other selectors: arithmetic Python can do is not left to a prompt. The job reads the store and writes only the wiki, so it stays inside one data domain, and it runs at 01:00 ahead of repair and consolidation, because writing last would leave every new page unchecked until the next night. `MEMORY_WINDOW_DAYS` moves the window through the same bounded validator `retention.py` uses.

It is deliberately narrow. People and attention pages only: a project needs a bounded outcome, a durable owner and a distinct identity, and one window of message traffic is weak evidence for all three; goals come from the person, and inferring them from an inbox is the overreach this design avoids. It also declines to invent a priority — when the evidence supports none it writes the page saying so, because a guessed priority makes the ranking job promote work the person never picked.

SOUL.md gains the two paths it never had. Asked what to focus on, the agent read `memory/index.md`, found nothing and reported that it knew of no priorities — with obligations sitting in the store. Asked for its recent Slack messages, it grepped the memory for "slack" and answered that it had no record of any, with 559 rows in `items`. It now names both tables, with the columns and a query that resolves the path in the same code, because `$HERMES_HOME` does not expand inside a file-reading tool and hunting for the database cost 26 tool calls and two minutes before the agent gave up.

This is independent of #130 and #137 and touches no file either of them changes except `README.md` and `register-jobs.sh`, where the additions are separate lines.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — 524 files
- [x] `git diff --check`
- [x] Recipe suite: 505 tests, `OK`
- [x] Run end to end on a real mailbox and Slack workspace under Hermes 0.19.0: the job wrote six people pages and both attention pages from 952 collected messages, and the repair job's own log shows what it then had to correct
- [x] The skill was tightened after that first run emitted structurally incomplete pages, and re-run against the same data: zero missing required fields straight from the writer
- [x] Every new assertion checked by breaking the behaviour it describes and confirming it turns red — the two-message threshold, the automated-sender rule, both bounds, the no-page-first ordering, decay detection, and the wake gate

- Result: `docs-updated`
- Evidence or justification: `README.md` — new job in the schedule table, `select_memory.py` in the file map, skill count, and a section on where the memory comes from and why the job is load-bearing; `profile/SOUL.md` — the store paths the agent needs to answer from what it holds.
- Reviewer: @rebelle5868

## Release And Compliance

- [x] No secrets or credentials are included.

### DIFF
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -238,7 +238,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the ten files report 270 tests in
+Expected result: every file ends with `OK`, the ten files report 280 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -247,7 +247,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the twelve files report 477 tests in
+Expected result: every file ends with `OK`, the twelve files report 482 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -88,9 +88,11 @@ those two apart.
 | `profile/scripts/walkthrough.py` | The fixture walkthrough, end to end, with no credentials and no model |
 | `profile/scripts/select_intake.py` | The intake job's pre-step: what to judge, and whether to wake the model at all |
 | `profile/scripts/select_review.py` | The review job's pre-step: the stalest open obligations, oldest review first |
+| `profile/scripts/select_memory.py` | The memory-writing job's pre-step: recurring correspondents, and whether the attention pages have gone stale |
+| `profile/seed/` | The pages a fresh memory starts from — copied in if missing, never overwritten |
 | `scripts/install.sh` | Installs the profile, inherits the runtime's model config, registers the jobs |
-| `scripts/register-jobs.sh` | Registers the five scheduled jobs through the cron CLI. Re-runnable |
-| `profile/skills/` | Five skills: judging, review, repair, consolidation, preference update |
+| `scripts/register-jobs.sh` | Registers the six scheduled jobs through the cron CLI. Re-runnable |
+| `profile/skills/` | Six skills: judging, review, memory writing, repair, consolidation, preference update |
 | `fixtures/` | Eight synthetic messages, a seed memory, and one recorded model turn |
 
 A later phase adds the optional Microsoft Graph and Slack connectors. Until
@@ -102,7 +104,7 @@ then the scheduled jobs judge and re-judge whatever the store already holds.
 - The fixture path — everything under [Try it](#try-it) and
   [Verify](#verify) — runs on Linux, macOS, or Windows under Windows Subsystem
   for Linux. Every command is written for a POSIX shell.
-- **The scheduled path is Linux only**, including WSL. All five shipped skills
+- **The scheduled path is Linux only**, including WSL. All six shipped skills
   declare `platforms: [linux]`, and Hermes refuses to load a skill outside its
   declared platforms. On macOS the jobs would still fire and the model would
   still be called — with no skill attached and a "skill not found" notice in
@@ -236,7 +238,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the nine files report 223 tests in
+Expected result: every file ends with `OK`, the ten files report 270 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -309,22 +311,55 @@ profile afterwards, so a setting that could not be written — or that reports
 success without sticking — ends the run rather than leaving a profile that
 took some of its configuration. The installer then checks both — that a model
 resolves and that a credential is present — and exits before registering any
-job if either is missing, rather than scheduling five jobs that would each
+job if either is missing, rather than scheduling six jobs that would each
 fail. If your endpoint genuinely needs no key, pass `ALLOW_NO_API_KEY=1` to
 say so.
 
 Re-running it is safe: the registration looks each job up by name and edits it
 rather than adding another copy.
 
-Five jobs are registered:
+Six jobs are registered:
 
 | Job | Schedule | Pre-step | Skill |
 | --- | --- | --- | --- |
 | intake | every 30 minutes | `select_intake.py` | `inbound-judging` |
 | review | every 6 hours | `select_review.py` | `obligation-review` |
+| memory writing | daily 01:00 | `select_memory.py` | `memory-writing` |
 | memory repair | daily 03:00 | — | `memory-repair` |
 | memory consolidation | daily 04:00 | — | `memory-consolidation` |
 | preference update | daily 04:30 | — | `preference-update` |
+
+**Where the memory comes from.** Three of the memory jobs maintain one and
+none of them creates it: repair checks invariants, consolidation compacts
+pages past their ceiling, preference-update writes the policy. The
+memory-writing job is what fills the gap. `select_memory.py` does the
+counting — who has been in touch inside the window, how often, who already has
+a page, which attention pages are past their decay window — and the skill
+decides who is worth a page and what it says. `MEMORY_WINDOW_DAYS` moves the
+window, which matters on a first run against a store that already holds
+months of history.
+
+That job is load-bearing rather than decorative. `ranking.py` reserves `high`
+for work the person has *chosen*, and only `attention/` and `goals/` can
+answer that. With an empty memory nothing reaches the top tier and the
+assistant is left measuring how loudly the outside world is asking, which is
+the thing it exists not to do. Measured on a real mailbox: 952 collected
+messages produced obligations that were all `medium` or `low`, because no page
+could gate one higher.
+
+It writes people and attention pages only, and `schema.md` now carries a
+table saying which page types have a writer at all — `projects/`,
+`patterns/`, `concepts/` and `event_triggers.md` do not yet, and a rule about
+a page nothing creates is a rule about a page somebody would keep by hand.
+Naming that is the point: the schema previously said ingest checked
+`event_triggers.md` against every incoming message, which no shipped job has
+ever done.
+
+`goals/` is the one absence that is a decision. It gates the ranking job's
+top tier alongside `attention/`, so a goal inferred from somebody's inbox
+promotes work they never chose. The job also declines to invent a priority:
+when the evidence supports none it writes the page saying so, because a
+guessed priority produces the same failure by the same route.
 
 **An idle tick costs nothing.** Each of the first two jobs runs its pre-step
 script first, then one agent turn over that script's output. When the script
@@ -364,8 +399,8 @@ provider takes over from the in-process ticker.
 **The job store is not part of the distribution.** `distribution.yaml` does not
 declare `cron`. An update replaces what it does declare — `SOUL.md`,
 `schema.md`, `skills`, `scripts` and the manifest — and leaves the jobs and
-their run history alone. Measured, not assumed: five jobs
-registered, `hermes profile update` run on Hermes 0.20.0, five jobs still
+their run history alone. Measured, not assumed: six jobs
+registered, `hermes profile update` run on Hermes 0.19.0, six jobs still
 there with the same ids. A test asserts
 the manifest never claims `cron` or `workspace`.
 
@@ -536,7 +571,7 @@ package, so nothing is added to the repository's third-party notices.
 The recipe's own scripts reach no network. They read the recipe's files from
 the checkout and write application state only inside the profile home, and they
 also leave a Python bytecode cache beside the scripts — see
-[Cleanup](#cleanup). The five skill files a runtime loads add nothing to that.
+[Cleanup](#cleanup). The six skill files a runtime loads add nothing to that.
 
 The scheduled path does reach one. A job that wakes runs an agent turn, and the
 runtime calls whichever inference provider it is configured for, over its own
@@ -560,7 +595,7 @@ Three separate questions, with three different answers.
 **Do the jobs survive?** Yes. They live in `$HERMES_HOME/cron/jobs.json`, which
 is ordinary disk state — not part of the distribution, so a profile update
 leaves it alone, and not tied to any process, so shutting everything down and
-starting again finds the same five jobs with the same ids.
+starting again finds the same six jobs with the same ids.
 
 **Do they start firing again on their own?** Only if the gateway was installed
 as a service. `hermes -p <profile> gateway install` registers one — a launchd
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -245,7 +245,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the twelve files report 494 tests in
+Expected result: every file ends with `OK`, the twelve files report 505 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -471,6 +471,18 @@ purpose: nothing matches on it but the memory job, and it is one column both
 sources agree on rather than a per-source pair. An excluded sender's identity
 is not stored either, because exclusion is applied inside `insert_items` and
 nothing about them is written at all.
+
+**An upgraded store does not get this retroactively.** The value was never
+kept, and it cannot be recovered from a display name — deriving it that way is
+the mistake the field exists to prevent. So rows collected before the upgrade
+carry no identity, and the memory job does not guess at one: it counts them,
+reports the count as `messages_without_identity`, and writes nobody's page
+from them. Two things close the gap. Re-reading a message fills its identity
+in, so a rolling collection window keys the recent past within a window; and
+the job only looks thirty days back, so the gap is gone once the window has
+turned over. What is lost in between is that a person's pre-upgrade messages
+do not count toward whether they are worth a page — not that anything was
+attributed to the wrong person, which is the failure this refuses.
 
 Four controls over what is kept ship alongside it, and they were in place
 before the collector that fills the store was. They apply to real Slack
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -100,12 +100,10 @@ those two apart.
 | `scripts/install.sh` | Installs the profile, inherits the runtime's model config, registers the jobs |
 | `scripts/register-jobs.sh` | Registers the seven scheduled jobs through the cron CLI. Re-runnable |
 | `profile/skills/` | Six skills: judging, review, memory writing, repair, consolidation, preference update. Retention needs none — it clears bodies and never wakes the agent |
->>>>>>> upstream/main
 | `fixtures/` | Eight synthetic messages, a seed memory, and one recorded model turn |
 
-Slack is connected through `scripts/setup-slack.sh`; a Microsoft Graph
-connector is still to come. Until a connector is set up, the scheduled jobs
-judge and re-judge whatever the store already holds.
+Slack is connected through `scripts/setup-slack.sh`. Until a connector is set
+up, the scheduled jobs judge and re-judge whatever the store already holds.
 
 ## Requirements
 
@@ -247,7 +245,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the twelve files report 484 tests in
+Expected result: every file ends with `OK`, the twelve files report 494 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -263,6 +261,7 @@ leave the loop exiting `0`.
 | Writer behavior, audit trail, caps across batches, correction idempotency, correction state transitions, displaced-row audit | `tests/test_apply_decisions.py` |
 | The walkthrough, and its central claims | `tests/test_walkthrough.py` |
 | Selector output, the wake gate, and the scheduler contract | `tests/test_selectors.py` |
+| What the memory job hands the agent: who is worth a page, stable identity across namesakes and renames, quiet days costing nothing, and corrections counted once | `tests/test_select_memory.py` |
 | Retention, exclusion, export and reset | `tests/test_lifecycle.py` |
 | The Slack collector: watermarks, partial failure, scope probing, thread discovery and rotation, the credential never reaching a stream, and the lifecycle controls applying to what it writes | `tests/test_ingest_slack.py` |
 
@@ -462,6 +461,16 @@ recipient lists are never stored. Ingest reduces them to a single `addressing`
 value — `direct`, `mentioned`, or `broadcast` — so the store never holds a
 copy of who else was on a thread. `normalize.py` does this today, and
 `tests/test_normalize.py` asserts it.
+
+One thing the store does keep, and should be said plainly rather than found
+in the schema: **each row carries the sender's stable identity** — their mail
+address, or their Slack user id — in `sender_key`. It is there because a
+display name cannot identify anybody, and a memory page named after one is
+overwritten by the next person who shares it. It is not kept for any other
+purpose: nothing matches on it but the memory job, and it is one column both
+sources agree on rather than a per-source pair. An excluded sender's identity
+is not stored either, because exclusion is applied inside `insert_items` and
+nothing about them is written at all.
 
 Four controls over what is kept ship alongside it, and they were in place
 before the collector that fills the store was. They apply to real Slack
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -247,7 +247,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the twelve files report 482 tests in
+Expected result: every file ends with `OK`, the twelve files report 484 tests in
 total, and the last line is `failed=0`. Do not use `|| break` here; a `for`
 loop reports the status of its last command, so a failing test would still
 leave the loop exiting `0`.
@@ -324,14 +324,14 @@ profile afterwards, so a setting that could not be written — or that reports
 success without sticking — ends the run rather than leaving a profile that
 took some of its configuration. The installer then checks both — that a model
 resolves and that a credential is present — and exits before registering any
-job if either is missing, rather than scheduling six jobs that would each
+job if either is missing, rather than scheduling seven jobs that would each
 fail. If your endpoint genuinely needs no key, pass `ALLOW_NO_API_KEY=1` to
 say so.
 
 Re-running it is safe: the registration looks each job up by name and edits it
 rather than adding another copy.
 
-Six jobs are registered:
+Seven jobs are registered:
 
 | Job | Schedule | Pre-step | Skill |
 | --- | --- | --- | --- |
@@ -705,7 +705,7 @@ Three separate questions, with three different answers.
 **Do the jobs survive?** Yes. They live in `$HERMES_HOME/cron/jobs.json`, which
 is ordinary disk state — not part of the distribution, so a profile update
 leaves it alone, and not tied to any process, so shutting everything down and
-starting again finds the same six jobs with the same ids.
+starting again finds the same seven jobs with the same ids.
 
 **Do they start firing again on their own?** Only if the gateway was installed
 as a service. `hermes -p <profile> gateway install` registers one — a launchd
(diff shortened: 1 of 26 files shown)


examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,246 @@
...
+        "interactions": found["interactions"],
+        "open_obligations": obligations,
+        "attention": stale_attention(),
+    }
+    print(json.dumps(report, indent=2, ensure_ascii=False))
+
+    # Nothing to write is the common case on a quiet day, and it must be free.
+    # A missing or stale attention page counts as work even when no person
+    # qualifies, because that page is what the ranking job gates on.
+    work = bool(found["people"]) or bool(report["attention"])
```
apurvvkumaria: Existing correspondents keep this job awake even when their people pages and both attention pages are current. `evidence()` retains every sender above the threshold, including `has_page: true`, so two unchanged messages keep `work` true for the full window. I reproduced this with a current people page and current attention pages. Please surface an existing person only when there is deterministic new evidence after the page's recorded interaction, and add a regression test in which unchanged messages end with `{"wakeAgent": false}`.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,246 @@
...
+             "addressing": addressing, "text": " ".join((text or "").split())}
+            for source, event_at, addressing, text in rows]
+
+    return {"people": chosen, "interactions": interactions}
+
+
+def open_obligations(conn) -> list[dict[str, object]]:
+    """What the assistant currently believes is owed, for the attention pass."""
+    rows = conn.execute(
+        "SELECT global_rank, priority, title, source_id FROM obligations"
```
apurvvkumaria, line 467: This payload does not contain evidence that the user chose any work. Rank, priority, title, and source ID are judgments derived from inbound requests. The memory-writing skill correctly forbids treating an external deadline or busy thread as a chosen priority, so a fresh installation can only write an empty `current_priorities.md`; the high tier remains empty. Please add a trustworthy user-originated signal, such as an explicit correction or commitment, or narrow the PR's claim and scope so it does not state that this closes the ranking-gate gap.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,246 @@
...
+    for sender in wanted:
+        rows = conn.execute(
+            "SELECT source, event_at, addressing,"
+            "       substr(COALESCE(subject, body), 1, 200)"
+            "  FROM items WHERE sender = ? AND event_at >= ?"
+            "  ORDER BY event_at DESC LIMIT ?",
+            (sender, since, MAX_INTERACTIONS)).fetchall()
+        interactions[sender] = [
+            {"when": (event_at or "")[:10], "source": source,
+             "addressing": addressing, "text": " ".join((text or "").split())}
```
apurvvkumaria, line 450: This value contains externally authored message text and is passed to an agent that writes durable memory. Please define the trust boundary explicitly: message content is evidence only and instructions embedded in it must never be followed. Prefer a constrained or validated write path where practical, and document the protection against an external message becoming a persistent instruction or invented priority.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def memory_root():
+    from _db import ledger_path
+    return ledger_path().parent.parent / "memory"
+
+
+def slug(name: str) -> str:
+    """`Dana Okoro` -> `dana_okoro`, per the schema's filename rule."""
+    cleaned = re.sub(r"[^a-z0-9]+", "_", (name or "").strip().lower())
```
apurvvkumaria: This mapping is not safe as an identity key. Non-Latin names such as `李明` produce an empty slug, while distinct names such as `A-B` and `A B` both produce `a_b`. That can overwrite or misattribute a person's durable page. Please use a stable collision-safe identity, or fail closed on empty and duplicate slugs, and cover both cases with tests.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/skills/memory-writing/SKILL.md
```diff
@@ -0,0 +1,152 @@
...
+It does not decide who deserves a page. That is judgment, and it is yours.
+
+## Read first
+
+`$HERMES_HOME/schema.md` is authoritative for page types, required
+frontmatter, section order, and growth ceilings. Read it before writing
+anything. A page that violates it is a defect the repair job will rewrite, so
+writing one costs two turns and gains nothing.
+
+Then read `$HERMES_HOME/workspace/memory/index.md` and any existing page you
```
apurvvkumaria, line 68: A clean profile installation does not contain `workspace/memory/`, `index.md`, `log.md`, or the people and attention directories. The first scheduled run therefore starts by reading and appending to paths that do not exist. Please add a deterministic bootstrap path and a clean-install test that produces the required directory structure, a valid index and log, and memory that passes the shipped invariant check. Do not rely on the model to guess the initialization contract.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,246 @@
...
+        if not path.exists():
+            found.append({"page": name, "state": "missing"})
+            continue
+        head = path.read_text(encoding="utf-8")[:400]
+        updated = re.search(r"^updated:\s*(\d{4}-\d{2}-\d{2})", head, re.M)
+        decay = re.search(r"^decay:\s*(\w+)", head, re.M)
+        if not updated:
+            found.append({"page": name, "state": "no updated field"})
+            continue
+        age = (today - datetime.strptime(updated.group(1), "%Y-%m-%d").date()).days
```
apurvvkumaria: A date with the expected shape but an invalid value, such as `2026-99-99`, raises here and aborts the memory-writing job before it can report the page. Please handle malformed and unreadable frontmatter as a finding instead of crashing, and add a negative test. This is especially important because the new job runs before the repair job.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -88,9 +88,10 @@ those two apart.
 | `profile/scripts/walkthrough.py` | The fixture walkthrough, end to end, with no credentials and no model |
 | `profile/scripts/select_intake.py` | The intake job's pre-step: what to judge, and whether to wake the model at all |
 | `profile/scripts/select_review.py` | The review job's pre-step: the stalest open obligations, oldest review first |
+| `profile/scripts/select_memory.py` | The memory-writing job's pre-step: recurring correspondents, and whether the attention pages have gone stale |
 | `scripts/install.sh` | Installs the profile, inherits the runtime's model config, registers the jobs |
-| `scripts/register-jobs.sh` | Registers the five scheduled jobs through the cron CLI. Re-runnable |
-| `profile/skills/` | Five skills: judging, review, repair, consolidation, preference update |
+| `scripts/register-jobs.sh` | Registers the six scheduled jobs through the cron CLI. Re-runnable |
+| `profile/skills/` | Six skills: judging, review, memory writing, repair, consolidation, preference update |
```
apurvvkumaria: Please update the remaining job and skill counts in the same change. This README still says five at the platform, installer, update-survival, sandbox, and reboot sections, and the installer and registration-script comments also say five. In addition, `select_memory.py` and the PR description cite `retention.py`, but that file is not present. The public documentation and implementation should consistently describe the six shipped jobs and skills without referencing an unavailable file.

apurvvkumaria (changes requested):
Thank you for adding the missing memory-writing path. All current GitHub checks are green, and the 243-test recipe suite passes locally. The remaining findings concern the core intent signal, idle-cost behavior, safe persistence of external text, first-run initialization, identity integrity, and documentation consistency. Please address the inline comments before approval.

apurvvkumaria (changes requested):
Thank you for the substantial update. The deterministic bootstrap, malformed-date handling, unchanged-correspondent gate, job counts, and documented trust boundary are improved. All 263 recipe tests and the required GitHub checks pass. Three blockers remain before approval:

1. **Read the real correction payload and distinguish positive from negative choices.** `correct.py` records a priority override as `after_json={"manual_priority": tier}`, but `select_memory.py` reads only `priority` or `status`. I reproduced a real `correct.set_priority(..., "low")` event reaching the selector as `"to": null`. The new test inserts `{"priority": "high"}` directly, so it does not exercise the production contract. Drive the regression through `correct.py`, preserve the actual target tier, and ensure only evidence that really means chosen work can inform `current_priorities.md`.

2. **Do not replay unchanged corrections as new work.** `corrections()` selects every user event inside the 30-day window, and `bool(chosen)` wakes the agent every day. I reproduced two consecutive runs with current attention pages and no state change; both omitted the wake gate. Add a durable processed watermark or another deterministic comparison so a correction wakes the writer once and becomes work again only when the underlying evidence or required page state changes.

3. **Use a stable person identity rather than a transformed display name.** The new slug avoids empty ASCII reductions, but it is not collision-safe as a durable identity. Two people with the same display name still share a page, while canonically equivalent spellings can produce different pages; for example, composed and decomposed forms of `Ünal Öz` generate different filenames. Preserve a source identity/address in the store and derive the page key from it, or fail closed when identity is ambiguous, with regressions for duplicate display names and Unicode normalization.

Non-blocking: update the pull-request description from 243 to 263 tests after the code changes settle.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,428 @@
...
+        " ORDER BY e.ts DESC LIMIT 40", (since,)).fetchall()
+    out: list[dict[str, object]] = []
+    for ts, event_type, after_json, title, source_id in rows:
+        try:
+            after = json.loads(after_json) if after_json else {}
+        except (TypeError, json.JSONDecodeError):
+            after = {}
+        out.append({"when": (ts or "")[:10], "action": event_type,
+                    "title": title, "source_id": source_id,
+                    "to": after.get("priority") or after.get("status")})
```
apurvvkumaria: This still reads keys that the production correction path does not write. `correct.set_priority()` records `after_json={"manual_priority": tier}`, so a real priority override reaches this report as `to: null`; the test currently inserts a synthetic `priority` key and masks the mismatch. Please drive the regression through `correct.set_priority()`, read `manual_priority`, and distinguish a positive selection from `low`/`ignored` evidence so only work the user actually chose can populate `current_priorities.md`.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,428 @@
...
+    this matters to them, and ignoring something is them saying it does not.
+    Both are choices, made deliberately, recorded with their subject.
+
+    Without this the job could only ever write an empty priorities page, and
+    the ranking gate it exists to feed would stay shut on every install.
+    """
+    rows = conn.execute(
+        "SELECT e.ts, e.event_type, e.after_json, o.title, o.source_id"
+        "  FROM events e JOIN obligations o ON o.id = e.obligation_id"
+        " WHERE e.actor = 'user' AND e.ts >= ?"
```
apurvvkumaria, line 501: Every user event inside the 30-day window is still returned on every scheduled pass, and `bool(chosen)` therefore wakes the agent again with no new evidence. I reproduced two consecutive runs with current attention pages: both surfaced the same correction and neither emitted the sleep gate. Please add durable, crash-safe application tracking—such as an event ID recorded in the successfully written page or an acknowledgement after the write—so an unchanged correction wakes once and is reconsidered only when its evidence or required page state changes.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,428 @@
...
+                         f"- seeded {', '.join(written)}\n")
+    return written
+
+
+def memory_root():
+    from _db import ledger_path
+    return ledger_path().parent.parent / "memory"
+
+
+def slug(name: str) -> str:
```
apurvvkumaria, line 160: The transformed display name is still not a durable person identity. Two people with the same display name map to one page, while canonically equivalent Unicode spellings map to different pages (`Ünal` versus `U\u0308nal`). The current normalization path also drops the email address and Slack user ID before storage. Please preserve a stable source identity/address and derive the key from it, or fail closed when identity is ambiguous, with regressions for duplicate display names and NFC/NFD-equivalent spellings.

apurvvkumaria (commented):
I re-reviewed the current head. The new seed pages and writer-scope documentation are useful, and the 270-test branch suite passes locally. The three functional blockers below remain. The branch also conflicts with current `main` in `README.md` and `profile/scripts/tests/test_selectors.py`, and GitHub has not reported checks for this head. Please resolve the functional findings first, merge current `main` normally, update the documented test count, and rerun the required checks.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Reproduced, and it is the third time in this recipe that I have written a fixture in a shape the writer never produces. `correct.set_priority()` records `{"manual_priority": tier}`; this read `priority`, so every real override arrived as `to: null` — and the test inserted a synthetic `priority` key, so the code and the test agreed with each other while neither agreed with `correct.py`.

The regressions go through `correct.set_priority()` and `correct.ignore()` now, and the selector reads `manual_priority`. Each correction carries a `direction`: raising to `high` is `chose`, a lower tier or an ignore is `declined`, and only `chose` may inform `current_priorities.md` — writing a declined one there would promote exactly what the person pushed away.

Reading it again turned up a third case I had classified as neither. `correct.py` emits three user events and this handled two: restoring something previously ignored is the person changing their mind and saying it is their work after all, which is as clear a statement as raising it to `high`, and it was being dropped. A test now fails if `correct.py` grows a fourth event this does not classify.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Reproduced. One thing the user did once cost a model call a night for the length of the window — the same defect as offering a correspondent whose page is already current, arriving from the other side.

I took the first of your two suggestions. The page records which corrections it accounts for as `<!-- applied: N -->` markers carrying the event id; the selector reads them back and hands over `unapplied_corrections` separately, and only that counts as work. The markers live in the page rather than in a file beside it because a separate record can be written when the page was not, or lost when the page was kept — in the page it is durable exactly when the page is, which is the only moment that matters. The skill says to write one for every correction used *and* every one deliberately declined, so a `low` override does not come back forever.

Two things I checked rather than assumed: `memory_check.py` reads a page carrying those markers as clean, and neither the repair nor the consolidation skill rewrites attention pages wholesale, so nothing else strips them.

The same reading turned up a related defect you did not see, because it needed more corrections than a test usually has. The forty-row bound was applied in SQL as `ORDER BY ts DESC LIMIT 40`, which drops the oldest — and the oldest are the ones most likely never to have been written up, so an unapplied correction could be pushed out of the window forever by newer traffic with nothing saying so. Unapplied ones are now never the part dropped, and a truncated pass reports `corrections_not_shown` rather than looking complete.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Agreed on both halves, and I took the second of your two options because the first needs something this PR does not have. `items` keeps the display name and drops the address and the Slack user id at normalization, so there is no stable identity in the store to derive a key from; adding one is a schema change, and it belongs with the connectors that would populate it rather than with the job that reads them.

So: NFC normalization first, which makes the composed and decomposed spellings of one name one page, with a test asserting the two differ as strings and agree as keys. And the ambiguous case fails closed — two senders that reduce to one page name are reported in `ambiguous_identity`, neither is written, and the rest of the pass continues. Silently merging them would put one person's history on the other's page under the other's name, and nothing in the store could separate them afterwards.

One more from the same area, found while testing this rather than from the review: a page whose `last_interaction` is in the future hid its person permanently. A typo or a clock skew is enough to write one, and the comparison then skipped that correspondent silently for as long as the date stood. A future date now reads as unknown, which means look.

rebelle5868:
All three are addressed inline, along with three more that came out of reading the same code again rather than from the review — a `restored` correction being classified as neither positive nor negative, the correction bound dropping the oldest silently, and a future-dated page hiding its person permanently.

Merged current `main`. #137 and #130 landed while this was in review, so every conflict was two sides adding a line to the same table and wanted the union: the file map carries `select_memory.py` and `seed/` alongside `retention.py`, `exclusions.py`, `export_store.py` and `reset.py`; the schedule carries memory writing at 01:00 and retention at 02:00; the guard in `test_selectors.py` expects both. Two places took main's wording instead of merging it — the note that `profile update` lists `cron/` among the directories it overwrites, and the sandbox section's description of the collector's egress, which is more accurate than this branch's "the recipe's own scripts reach no network".

Tests: 482 across twelve files, up from 243 across ten when this opened. Checks are reporting again; the earlier gap was my own rapid force-pushes rather than anything about the head.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
apurvvkumaria: The real-path coverage is improved, but `restored` is still misclassified as `chose`. I reproduced `set_priority("m1", "low") -> ignore("m1") -> unignore("m1")`: the obligation remains `manual_priority=low` and `priority=low`, while the restored event is emitted as `direction: chose`. The skill may then promote work the user explicitly kept low. Restoring means reopen or track again; it does not establish top-priority intent. Please classify only an explicit high override as `chose` (or add a distinct persisted signal for that intent), and add this low -> ignore -> restore regression.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
apurvvkumaria: The crash-safe marker fixes the repeated wake. One truncation claim still needs correction: with 45 unapplied events, the current code returns 40 and `corrections_not_shown: 5`, so five unapplied corrections were not provided. That directly conflicts with both “never at the cost of losing an unapplied choice” and the skill text “Everything unapplied is always in what you were given.” The batching behavior can be valid because later runs can drain the remainder, but please describe it accurately and test that the omitted events are surfaced on a subsequent pass after the first batch is acknowledged.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
apurvvkumaria: After the merge from current `main`, the README now has seven rows in the job table, but the surrounding setup text still says “scheduling six jobs,” “Six jobs are registered,” and “the same six jobs” after reboot (lines 327, 334, and 708). Please change those job counts to seven. The six-skill references are accurate and should remain six.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
apurvvkumaria: The identity blocker remains. The new test checks only that NFC/NFD forms produce the same slug; it does not send both through `evidence()`. Doing so reports the canonically equivalent spellings as ambiguous because `by_slug` stores the two raw strings. Also, four messages from two sources with the exact display name `Alex Kim` produce one person with four mixed interactions and no `ambiguous_identity`, because the store has already discarded the stable IDs. Please add end-to-end regressions for both cases. Since exact duplicate names are not observable after normalization, preserving a stable source identity (or narrowing the feature so it cannot write a conflated person page) is still required.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Reproduced exactly as you described: `set_priority("m1","low")`, `ignore`, `unignore` leaves the obligation at `manual_priority=low` while the restore arrived as `direction: chose`.

Only an explicit `high` override is choosing now. Restoring means track this again, which is a real thing the person did and worth recording on the relevant page, but it does not establish priority — and reading it as one would promote work they had deliberately kept down, through the single event that looks like agreement.

The regression asserts the fixture really does leave the obligation at `low` before checking the direction, so it cannot pass on a setup that fails to reproduce the sequence.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: You are right, and the code was defensible while the description was not. With more unapplied corrections than the bound, unapplied ones were dropped — contradicting both the comment and the skill's "everything unapplied is always in what you were given".

Both now describe what it is: a batch, unapplied first, with the remainder arriving on a later pass once this batch's markers are on the page. The test drives two passes and asserts the remainder arrives, that nothing acknowledged is re-offered, and that the union is the whole set — which is the property that makes the batching acceptable rather than lossy.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: All three — the installer paragraph, the registration line and the reboot section. The six-skill references are unchanged, as you said.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Both halves reproduce, and the first one means my fix made things worse rather than incomplete. Sent through `evidence()`, the composed and decomposed spellings arrive as two distinct raw strings, land on one slug, and are reported as ambiguous — so neither page is written. I turned a silent merge into a double refusal, and the test could not see it because it called `slug()` directly rather than going through the selector.

The second half I also confirmed: four messages from two sources with the display name `Alex Kim` produce one person with four mixed interactions and nothing in `ambiguous_identity`, because the strings are identical and the stable ids were discarded at normalization.

Preserving a source identity is the right answer, and it is a schema change — `items` keeps the display name and drops the address and the Slack user id, though both normalizers already compute them. Before I make it, one question of ordering: #142 currently takes schema v3 for its tombstone column, and I would rather agree the numbering with you than have two open PRs mint the same version. I am happy to take v4 here, or to move #142 up — it is the younger PR and unmerged. Which would you prefer?

apurvvkumaria: Use schema v3 for the stable source identity in #140, then rebase #142 and move its tombstone migration to schema v4. That keeps the older independent PR from depending on the younger one and gives us a sequential v2 -> v3 -> v4 migration path. Please include end-to-end regressions for NFC/NFD-equivalent names and identical display names from different source identities.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -91,13 +91,16 @@ those two apart.
...
 | `profile/scripts/exclusions.py` | Senders, domains and channels that are never written, applied in `insert_items` |
 | `profile/scripts/export_store.py` | Writes the whole store and memory out as Markdown beside JSON |
 | `profile/scripts/reset.py` | Removes the store, the memory and the learned policy together |
+| `profile/seed/` | The pages a fresh memory starts from — copied in if missing, never overwritten |
 | `scripts/install.sh` | Installs the profile, inherits the runtime's model config, registers the jobs |
-| `scripts/register-jobs.sh` | Registers the six scheduled jobs through the cron CLI. Re-runnable |
-| `profile/skills/` | Five skills: judging, review, repair, consolidation, preference update. Retention needs none — it clears bodies and never wakes the agent |
+| `scripts/register-jobs.sh` | Registers the seven scheduled jobs through the cron CLI. Re-runnable |
+| `profile/skills/` | Six skills: judging, review, memory writing, repair, consolidation, preference update. Retention needs none — it clears bodies and never wakes the agent |
+>>>>>>> upstream/main
```
apurvvkumaria: **Blocker:** This README still contains the merge marker `>>>>>>> upstream/main`. Please remove it and run `git diff --check`; that check currently fails at this line.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -0,0 +1,622 @@
...
+    sandbox that cannot create one, that surfaces as `unable to open database
+    file` — an error that reads like a permission or locking fault and is
+    neither. A fixture-sized store never reaches the spill, so the bug is
+    invisible until the first real mailbox. Keep the payload small here.
+    """
+    # Grouped by identity and display name together, so a person who changed
+    # how their name is shown is still one person, and two people who share a
+    # display name are still two.
+    counted = conn.execute(
+        "SELECT COALESCE(sender_key, sender), sender, COUNT(*), MAX(event_at)"
```
apurvvkumaria: Blocker: Falling back to the display name makes the v2→v3 upgrade split one real person into two identities. The migration intentionally leaves existing rows with `sender_key = NULL`, and `insert_items()` uses `INSERT OR IGNORE`, so re-reading an old `source_id` with its now-known key does not backfill it. I reproduced two legacy “Dana Okoro” rows plus two newly keyed rows producing two candidates and two page slugs: one keyed by `Dana Okoro`, one by `dana@example.com`. Please either provide an explicit safe backfill/recollection path or fail closed on unkeyed legacy rows instead of treating a display name as identity. Add an end-to-end v2 migration regression through normalization and `evidence()`; the public README’s statement that each row carries a stable identity also needs to describe the actual upgrade boundary.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/schema.md
```diff
@@ -38,13 +38,40 @@ An append-only note store needs no schema, and gets none of these properties:
...
+`current_priorities.md` is careful to avoid. Goals come from the person.
+
 ## Page types
 
 ### People (`people/`)
 
 ```yaml
 ---
 name: Full Name
+source_key: person@example.com     # the address or user id this page is about
```
apurvvkumaria, line 74: Blocker: `source_key` is now the load-bearing identity field, but the deterministic memory checker does not enforce it. `memory_check.py` still omits it from `REQUIRED_FIELDS["people"]` and has no duplicate-key invariant; I confirmed that two otherwise valid people pages with no `source_key` both pass `check_all()`. An agent omission or duplicated key can therefore orphan or conflate pages without the repair path noticing. Please require a non-empty key for newly written people pages, detect duplicate keys, define safe handling for legacy pages without inventing an identity, and add regressions for missing and duplicate values.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Reproduced, and it is worse than a fallback being imprecise — it was a display name being used as an identity again, in the one place I had just argued it could not be.

There are now two paths instead of a guess. `insert_items` fills in an identity a row does not have when a collector re-reads that same message, which `INSERT OR IGNORE` was discarding; the value comes from the message rather than from matching a name against another row, so it cannot merge two people the way a name-based backfill would. Whatever is still unkeyed is not attributed to anybody — those rows are counted and reported as `messages_without_identity`, and no page is written from them. The gap is bounded twice over, by the collection window and by the job's own thirty days.

The regression is end-to-end through `normalize` and `evidence()`, in your arrangement: two legacy rows and two keyed rows for one person now produce one candidate keyed by the address. Alongside it, a store with only legacy rows produces no people and a count, a re-read backfills without duplicating the row, and the backfill never overwrites an identity already present.

The README now describes the upgrade boundary rather than only the steady state, including what is actually lost in between: a person's pre-upgrade messages do not count toward whether they are worth a page. Nothing is attributed to the wrong person, which is the failure this refuses.

One change beyond the finding: the base test fixture now writes an identity. A row whose `sender_key` is a display name is a shape no normalizer produces, and a fixture in a shape production never produces is how three earlier defects here stayed invisible.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/schema.md
rebelle5868: You are right that the field the memory job matches on was the one thing the checker did not look at.

`check_identity` reports `missing-identity` and `duplicate-identity`, and is wired into `check_all` — with a test for that specifically, since a check that exists but is not called has never run.

Neither is repaired by guessing, and the repair skill now says so explicitly: a missing key is taken from the selector's `source_key` for that person or left absent and noted, and a duplicate is never resolved by merging or by picking, because two pages claiming one identity means one of them is about somebody else and only the selector's evidence can say which. "Invent a `source_key`, or copy one from another page to clear a finding" is now in the list of things repair must never do.

Regressions cover a missing key, a present-but-empty key, two pages claiming one key, and the duplicate being reported independently of directory order — `glob` order is filesystem order, and a finding that appears only on some machines is worse than no finding. The shipped fixture pages carry identities so the clean run is still clean for the right reason.

apurvvkumaria (approved):
Reviewed the latest head. The requested correctness, trust-boundary, bootstrap, stable-identity, migration, and verification issues are addressed. The complete 505-test recipe suite and required repository checks pass.

PR #141, opened 2026-08-26
feat(tools): add an agent memory benchmark with synthetic corpora

### LINKED ISSUE #136: feat(tools): add an agent memory benchmark with synthetic corpora
## Problem

Several examples in this repository build memory and then act on it — the Memory-Driven Chief of Staff recipe keeps a revisable local record per message, and the Developer Community Chief of Staff recipe synthesizes signals across sources. Both make a design choice about *when* to spend reasoning: consolidate at ingest time, or store cheaply and retrieve at question time.

There is currently no shared way to check what that choice bought. A contributor who changes a memory design can see that the example still runs, but not whether the memory it produces can actually answer questions about the material it read, and not what the choice cost in tokens. Comparisons between memory designs come down to reading each other's descriptions.

The Harness Engineering Playground covers a neighboring problem — it tunes a harness profile against agent tasks — but it measures harness behavior rather than the quality of memory built from a corpus.

## Proposal

We would like to contribute a standalone evaluation tool, and we would welcome maintainer guidance on placement before we open a pull request. Our reading of the example taxonomy is that this belongs under `examples/tools/`, as a developer and evaluation utility rather than a deployed agent blueprint, alongside the Harness Engineering Playground. A working name is `examples/tools/agent-memory-benchmark/`.

**What it does.** It models a general knowledge-work workflow: the email and chat one person accumulates across a few months of a project. It gives a system that corpus, lets it build whatever memory it likes, then asks 186 questions about the corpus and scores the answers. Quality and cost are reported separately and never blended: accuracy per question type on one axis, tokens spent building memory versus tokens spent per answer on the other.

**How it grades.** Deterministically, with no model in the scoring path. Answers are normalized for case, punctuation, and date spelling, then matched against an accept / reject / require-all rule set. Questions were written to have short, checkable answers so that the scoring path contains no model at all. This was a constraint rather than a discovery: a judge model gets deprecated and retuned, and each of those events silently re-scores history.

**How a system joins.** An adapter is a small JSON file naming two commands, one to ingest and one to answer, so any system can be scored without changing the harness. Token cost is counted by a local accounting proxy rather than self-reported.

**What the content is.** Both corpora are fully synthetic. Each one runs the same general workflow through a different domain, so that a result on one can be checked against the other:

| | Documents | Period | Domain |
| --- | ---: | --- | --- |
| Corpus A | 425 | six weeks | a software platform engineering team |
| Corpus B | 173 | ten weeks | a construction program manager |

Every person, company, project, domain, and email address is invented, and every message body was generated from a fixed fictional cast defined before any document existed. Nothing was collected from a live account. Corpus B was written key-first by a different model family, specifically so that no system's familiarity with the material could explain a result.

**Licensing and dependencies.** We would contribute everything under Apache-2.0 with the standard SPDX headers, matching the rest of the repository, so this introduces no license change and no third-party licensing question. Because the material is synthetic and authored for this purpose, there is no vendored third-party code, no model weights, and no binary in the contribution. The harness itself runs on the Python standard library alone; the retrieval baselines additionally call an OpenAI-compatible embeddings endpoint, also over the standard library. Any adapter for a system that needs more declares its own requirements separately.

**Size.** About 3.2 MiB, all plain text — the two corpora, the question set, the answer key, and the harness, as small Markdown and JSON files. We would rather ask than assume: if that is more than the repository wants to carry, we are happy to contribute Corpus A alone (1.8 MiB) and host the second corpus elsewhere, or to shrink either.

**Reference results.** We have run three systems through it on one base model — a consolidating design and two retrieval baselines. We would suggest keeping results out of the first pull request and proposing them separately, for two reasons. First, the benchmark's own methodology asks a valid submission to run at least two base models and publish the difference, and we have one so far. Second, the most useful thing the numbers produced is a caution rather than a ranking: the consolidating system's lead over retrieval was large on the corpus it was more familiar with and mostly disappeared on the corpus it had never seen. We would rather land the measuring instrument first and let others reproduce it than lead with our own scores.

## Alternatives

**Place it under the Memory-Driven Chief of Staff recipe, in `fixtures/`.** This was our first thought, since that recipe is the most obvious consumer. We set it aside for three reasons. That directory is documented as eight messages and a small seed memory whose purpose is to run the recipe end to end without a mailbox, a Slack workspace, or an inference endpoint — a benchmark corpus is a different kind of artifact with a different job. It is currently 44 KiB, so the corpora would grow it by two orders of magnitude and make the recipe heavier for everyone who only wants to run it, against the guidance that examples stay independently deployable. And the benchmark scores several systems, only one of which resembles that recipe, so filing it underneath would read as the recipe's self-test rather than a neutral instrument.

**Keep it outside this repository and link to it.** Simplest for us, but a contributor changing a memory design here could not run it or check a claim against it, which is most of the value.

**Contribute the corpora without the harness.** Smaller, but a corpus with no grader invites everyone to write their own scoring rules, and results stop being comparable — which is the problem we are trying to solve.

**Fold it into the Harness Engineering Playground.** Adjacent, but that tool optimizes a harness profile for a model, and this measures memory built from a corpus. Combining them would blur what each result means.

---

## Questions for maintainers

1. Is `examples/tools/agent-memory-benchmark/` the right placement, and does the name fit the naming policy? We are happy to rename.
2. Is roughly 3.2 MiB of plain text acceptable to carry, or would you prefer one corpus, or a smaller sample with the full set hosted elsewhere?
3. Would you like reference results in the repository at all, or should the tool ship with no scores and leave the leaderboard to whoever runs it?

rebelle5868:
## Related Issue

Closes #136

## Description

Adds a standalone evaluation tool that measures the memory a system builds from a corpus, rather than the harness around the model. It models a general knowledge-work workflow — the email and chat one person accumulates across a few months of a project — gives a system that corpus, lets it build whatever memory it likes, then asks 186 questions and scores the answers. Quality and cost are reported separately and never blended: accuracy per question type on one axis, tokens spent building memory versus tokens spent per answer on the other.

This follows the guidance on #136, including the follow-up clarification: it ships no model scores, and the offline oracle and intentionally-failing adapters are the acceptance controls.

**What is here**

| Path | What it is |
| --- | --- |
| `corpus/` | 425 documents in two dated halves, plus a manifest. The second half supersedes the first in places. |
| `corpus_b/` | 173 documents in an unrelated domain, written key-first by a different model family, so a result on one corpus can be checked against the other. |
| `questions/`, `gold/` | 186 questions and the answer key, with its per-mode grading rules. |
| `bench/` | Runner, grader, accounting proxy, price table, content fingerprinting. |
| `adapters/` | Two retrieval baselines, a ledger adapter over a memory store that ships in this repository, and the adapter contract. |
| `tools/` | Re-grade stored answers against the current rules. |
| `selftest/` | A six-document corpus and two adapters whose scores are known in advance. |
| `docs/` | Methodology, provenance, submission instructions. |
| `tests/` | 135 tests. |

**An incomplete submission is not a lower score.** A question a system produced no answer for is distinct from one it answered by declining: the first earns nothing and invalidates the run, the second is graded on its merits. Both the runner and `tools/regrade.py` record what was missing.

**Grading is deterministic.** No model sits in the scoring path. Answers are normalized for case, punctuation, and date spelling, then matched by a per-mode rule set. A judge model gets deprecated and retuned, and each of those events silently re-scores history, so questions were written to have short, checkable answers instead.

**The runner does not hand the answer key to the system under test.** Adapters are launched from a scratch directory rather than the benchmark root, so a relative `open("gold/answers.jsonl")` finds nothing, and each phase refuses to start if it was handed a path it must not have: ingest never receives the questions or the key, answer never receives the key. That is a guard against accidental leakage and not a sandbox — an adapter that goes looking can still derive the path, which the documentation states and a test demonstrates. See the known gaps below.

**Every phase is bounded.** A wall-clock budget per phase call, each adapter in its own process group, so a timeout or a Ctrl-C reaches the workers an adapter started rather than leaking them. Commands are argument arrays, never shell strings. The README states plainly that an adapter runs local code you chose to run and is not sandboxed.

**Cost is measured, not self-reported, and it fails closed.** The runner points `OPENAI_BASE_URL` / `ANTHROPIC_BASE_URL` at a local proxy and counts what crosses the wire. Every adapter declares whether it intends proxy accounting or hosts its own model, so silence can be read: an adapter that declares `proxy` and produces no traffic is an invalid run rather than a free one, and a forwarded request that returns no countable usage invalidates the run too, because a cost nobody measured must not read as a cost of zero. Only a fully counted run carries `comparable_on_cost: true`.

**A stored result is self-describing.** `report.json` carries `schema_version`, a fingerprint of the corpus, questions, answer key, date-normalization settings **and the scorer** — a hash of the modules that turn an answer into a verdict — plus the adapter revision, the benchmark repository revision, the declared model, and the trial index. Two rows are comparable only when their fingerprints match, and a report regraded under changed rules visibly stops matching one graded under the old ones. `tools/regrade.py` refreshes both the fingerprint and the benchmark revision, so a regraded report names the scorer that actually produced its verdicts.

No table renderer ships. This change carries no scores, no leaderboard is hosted anywhere, and a renderer that grouped rows on corpus and model alone would have ranked results with different question sets, answer keys and grading rules. Comparing runs is the reader's job; `docs/SUBMITTING.md` says what makes two reports comparable.

**A memory store that ships here can be driven end to end.** `adapters/ledger_rag/` loads the corpus into the Memory-Driven Chief of Staff recipe's own `schema.sql` and answers by selecting candidates out of that ledger. What it measures is candidate selection over the ledger plus the model — not the recipe's own behaviour, since that ledger is built for triage and ranking and has no question-answering path. The selection and the answer call belong to the adapter; nothing writes to or changes the recipe. The adapter, its `adapter.json` and the README all say so in those words. Its ingest is plain SQLite and needs no network, which is why the offline suite covers it.

**The corpora are fully synthetic.** Every person, company, project, domain, and address is invented, and every message body was generated from a fixed fictional cast defined before any document existed. Nothing was collected from a live account. Every address sits under the RFC 2606 reserved `.example` top-level domain, and a test fails if one does not. `docs/provenance.md` records the generation method, what was screened before publication, and content hashes; a test recomputes those hashes so the record cannot drift.

**Size.** 3.5 MiB across 664 files, all plain text. No third-party package is added — the harness and the baselines run on the Python standard library alone.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — all 561 checked source files have SPDX headers
- [ ] `python3 scripts/check_label_taxonomy.py --check` — not applicable; no governance metadata changed
- [x] `python3 scripts/check_pr_title.py "<title>"` — title follows the canonical convention
- [x] `git diff --check`
- [x] Example checks: `python3 -m pytest tests/` in the example directory — **135 passed**, offline, with no model, network, or API key

The suite is not structural. It runs the whole pipeline against `selftest/`, where an oracle adapter scores exactly 1.0 and a deliberately wrong adapter scores exactly 0.0 through a different code path per grading mode. It exercises the runtime guarantees by violating them: an adapter that reaches for the answer key, one that never finishes, and one that leaves a child process behind — reproducing the previous behaviour makes each of those fail. And it pins the claims documentation makes: the published content hashes, the stated test count, the difficulty split, and that no address uses a registrable domain.

| Test file | Tests | Covers |
| --- | ---: | --- |
| `test_answer_key_rules.py` | 57 | corpus A: id uniqueness, accept/reject separation, gold citations resolving, mode fields, determinism, published hashes, reserved domains, the stated test count, and that the answer contract names no real document |
| `test_grader.py` | 16 | grading rules |
| `test_corpus_b_rules.py` | 11 | corpus B held to the same invariants, including that no question cites a body-less document and that the documented counts match the shipped file |
| `test_accounting.py` | 9 | counted, missing-usage, bypassed, locally-hosted and mis-declared runs; only a fully counted one is cost-comparable |
| `test_ledger_adapter.py` | 9 | the ledger adapter offline: recipe schema present, corpus loaded through its constraints, ingest idempotent, selection finds cited evidence, and the answer phase running at one, three and six workers against a loopback stub |
| `test_missing_answers.py` | 11 | saying nothing earns no abstention credit — absent row, id-only row, blank answer — and invalidates the run in both the runner and regrade; a corrected submission clears the stale invalidation while an accounting failure survives it; an aggregation change moves the fingerprint |
| `test_runner_isolation.py` | 6 | the guards that do hold: a phase handed the key refuses to start, phases time out, child processes are cleaned up |
| `test_corpus_integrity.py` | 5 | corpus, questions and answer key agreeing |
| `test_selftest_end_to_end.py` | 5 | the whole pipeline offline, against a known score |
| `test_report_rendering.py` | 4 | no raw Python value in a rendered summary, on either path |
| `test_isolation_is_not_a_sandbox.py` | 2 | the guard that does not hold, demonstrated so it is a tested fact |

- Result: `docs-updated`
- Evidence or justification: `README.md`, `docs/methodology.md`, `docs/provenance.md`, `docs/SUBMITTING.md`, `corpus_b/README.md` and the adapter descriptions went through five review rounds, each one verifying the previous round's fixes against the files rather than against the commit messages — which caught three rounds where a claimed fix had not landed. The fifth round's findings are fixed in `fd8e4d9` and have not themselves been through another full round; the substantive ones are covered by tests added in that commit, and the outstanding items are listed below for the reviewer rather than silently closed. `examples/README.md` and the canonical taxonomy mapping each gain one row.
- Reviewer: Xuan Wu
- [x] Changed user-facing text follows the writing guide and controlled-word list
- [x] A public contributor can understand the changed text without internal company context
- [x] I reviewed any agent-generated text before submission

**Known gaps, stated rather than closed:**

- Verified on macOS and Python 3.9.6 only. The shipped adapter definitions invoke `python3`, which a default Windows install does not provide; the `## At A Glance` table says so.
- The artifact is called *mnemo* in its README and *Agent Memory Benchmark* in the catalog. One term would be better; say which you prefer and I will make it consistent.
- The answer key is reachable by an adapter that goes looking for it — `PYTHONPATH` carries the benchmark root. The documentation says so and `tests/test_isolation_is_not_a_sandbox.py` demonstrates it. Making it genuinely unreachable needs a filesystem boundary this harness does not have; happy to build one if you want it.

## Release And Compliance

- [x] No secrets or credentials are included, including API keys, access tokens, passwords, local `.env` files, private certificates, or token caches
- [x] No nonpublic project names, environment names, hostnames, URLs, ticket identifiers, workspace paths, logs, screenshots, or configuration values are included
- [x] Third-party dependency changes are reflected in `THIRD-PARTY-NOTICES` — no dependency is added, so no change is needed
- [x] Public content uses sanitized examples and placeholders instead of private values
- [x] I added my DCO sign-off declaration to this pull request description


apurvvkumaria (changes requested):
Thanks for this contribution. The required checks pass, and the complete local test suite passes with 115 tests. I found the following issues that need fixes before approval:

1. Make model-usage accounting fail closed. The proxy increases the call count only when a response contains countable usage. A streaming response without usage, an endpoint that omits usage, or an adapter that bypasses the proxy can therefore produce a valid-looking score with zero or unobserved cost. Track forwarded requests separately from countable usage, invalidate live runs with uncounted or bypassed calls, and exclude unobserved-cost rows from cost comparisons. Add regression tests, including equal-cost and missing-usage cases.

2. Enforce the documented answer-key boundary. The adapter receives the repository root through PYTHONPATH, so it can derive and read gold/answers.jsonl even though the command and environment do not contain the key's absolute path. The ingest phase can similarly reach the questions and answer key. Run untrusted adapter phases with a filesystem boundary that makes these files unavailable, or narrow the documentation and result claims so they do not assert isolation that the harness does not provide. Add a test that attempts indirect access.

3. Preserve and enforce result comparability. tools/regrade.py does not replace the fingerprint in an existing report after regrading against different inputs. tools/leaderboard.py groups only by corpus fingerprint and model, so it can compare rows with different question sets, answer keys, normalization rules, or benchmark revisions. Recompute the fingerprint during every regrade and group or reject comparisons using the full scoring fingerprint plus the applicable benchmark or grader revision.

4. Align the publication material. The catalog says corpus B has 96 questions, while the benchmark and its documentation contain 97. Correct the catalog entry and add the result screenshot required for a new example by CONTRIBUTING.md.

Please retain the existing DCO and SPDX compliance while updating the PR.

rebelle5868:
All four are addressed in `6285d79`. Three were straightforward; the second involved a choice I want to put in front of you rather than bury in a diff.

## 1. Accounting fails closed

Reproduced first: an adapter against an endpoint that returns no `usage` block made six real model calls and the report said `answer_calls: 0`, zero tokens, and `valid` unset. The proxy incremented its counter only inside the branch that found countable usage, so a forwarded request whose response omitted usage was not recorded at all — not the tokens, not the fact that a call happened.

The proxy now counts every request it forwards, separately from the ones it could count. A run whose forwarded count exceeds its counted one is marked invalid with the gap named, and the description says the cost is a floor rather than a total. A run where nothing crossed the proxy says so and names both readings — a locally-hosted model, or an adapter that ignored `OPENAI_BASE_URL`. `tests/test_accounting.py` covers the counted, missing-usage, no-traffic and equal-cost cases; reverting the fix turns the missing-usage test red.

## 2. The answer-key boundary — I narrowed the claim rather than building the sandbox

You are right, and it is worse than the description. I wrote a twenty-line adapter that imports `bench.runner`, reads `REPO` from it, opens `gold/answers.jsonl`, and answers out of the key. It scored 0.82 while `_assert_isolated` passed, the scratch working directory did its job, and every isolation test I had written stayed green. The guards check that the key is not *handed* to a phase; they do nothing about a phase that derives the path, and `PYTHONPATH` hands over everything needed to derive it.

I took the second of your two options: the README and the runner docstring now state that the key is reachable, that the guards prevent accidental leakage and not a determined adapter, and that a score is meaningful only for an adapter you trust. `tests/test_isolation_is_not_a_sandbox.py` demonstrates the bypass, so the limit is a tested fact rather than something a reader discovers in someone's result — and if a future change ever does close it, that test fails and points at the two documents that must stop saying it is open.

My reasoning for not building the boundary, which you should overrule if you disagree: this pull request ships no scores, there is no hosted leaderboard, and nothing here ranks anybody against anybody. Someone runs this against their own system for their own information, where an adapter that reads the key is only deceiving its author. A filesystem boundary would also be the example's first platform-specific dependency — `bubblewrap` or a container on Linux, `sandbox-exec` on macOS, neither on Windows — against a tool that currently runs on the standard library alone, and it would still not stop an adapter from reaching the network or caching answers into its own state directory between phases.

If you want the boundary anyway, say so and I will build it. It is a bounded piece of work; I did not want to add a platform dependency and a partial guarantee on my own initiative when the honest sentence was available.

## 3. Comparability

`tools/regrade.py` recomputes the fingerprint on every regrade. It previously left the old one in place, so regrading a report against a different question set produced a document whose accuracy and whose stated inputs disagreed.

For the second half I removed `tools/leaderboard.py` and `tools/compare_subset.py` rather than repairing them. The defect you found is real — grouping on corpus and model alone would have ranked rows with different question sets, answer keys, normalization rules and benchmark revisions, and it also emitted a cost conclusion about them. But this change ships no scores, no leaderboard is hosted anywhere, and neither the issue nor your guidance ever asked for one; the word came from a tool inherited from the internal version. A renderer with nothing to render was speculative infrastructure carrying a live bug. Comparing runs is now the reader's job, and `docs/SUBMITTING.md` states what makes two reports comparable. The leaderboard vocabulary is gone from the documentation and the module docstrings.

Happy to bring a comparison tool back later, with full-fingerprint grouping, alongside the reference results.

## 4. Publication material

The count was mine: dropping a corpus B question during review moved its total to 96, I corrected the catalog and left the submission guide at 97. Both now read 96, and a test pins each to the shipped question file so the next drop cannot leave two truths in the tree.

The README gains a `## Screenshot` section above `## At A Glance`. It is an SVG terminal frame — an image on GitHub, plain diffable text in the repository, no binary — showing the offline self-test scoring 1.0 on every question type and the suite passing, with the essential output preserved as searchable text beside it, per the template's note for command-line examples.

DCO and SPDX are unchanged. The suite is 119 tests, all offline.

apurvvkumaria (commented):
I re-reviewed the current head. The answer-key claim is now appropriately narrowed, missing-usage calls fail closed, the catalog count and screenshot are corrected, and the complete 119-test suite passes locally. All required GitHub checks are green and the branch is mergeable. The two functional blockers below remain. Non-blocking: please update the PR description from 115 to 119 tests and from 551 to the current 550 license-checked source files after the code settles.

examples/tools/agent-memory-benchmark/bench/runner.py
```diff
@@ -0,0 +1,467 @@
...
+        # only if all three hashes match; see docs/provenance.md.
+        "fingerprint": fingerprint(args.corpus, args.questions, args.gold),
+        "corpus": {
+            "documents": len(manifest),
+            "part_a": sum(1 for m in manifest if m["part"] == "part_a"),
+            "part_b": sum(1 for m in manifest if m["part"] == "part_b"),
+        },
+        "timing": timing,
+        "accounting": {
+            "method": ("partial" if uncounted else "proxy") if forwarded else "none-observed",
```
apurvvkumaria: The missing-usage case now fails closed, but the bypass case still produces a valid `none-observed` result. With `forwarded == 0`, the runner explicitly says the adapter may have bypassed the proxy, yet it never sets `valid: false`; that is the same unmeasured-cost path the prior review asked to exclude from cost comparisons. Please require the adapter to declare an accounting mode (for example, proxy-measured versus intentionally local/unmeasured), invalidate zero-traffic runs that claim proxy accounting, and add a regression showing a bypassed remote call cannot produce a valid cost row.

examples/tools/agent-memory-benchmark/tools/regrade.py
```diff
@@ -0,0 +1,103 @@
...
+                "part_b": sum(1 for m in manifest if m.get("part") == "part_b"),
+            }
+        else:
+            report["corpus"] = {"documents": 0, "part_a": 0, "part_b": 0}
+        before = None
+    report["summary"] = summarize(verdicts, gold)
+    # The fingerprint is the report's claim about what it was graded against.
+    # Regrading against a different question set or answer key changes that
+    # claim, so it is recomputed here rather than carried over.
+    report["fingerprint"] = fingerprint(args.corpus, args.questions, args.gold)
```
apurvvkumaria, line 92: Recomputing the input hashes fixes one half of the comparability defect, but this can still rewrite the score with the current grader while retaining an old `benchmark_revision` from the original report. The fingerprint itself hashes only corpus, questions, gold, and normalization settings—not the grader implementation—while the submission guide says matching fingerprints are sufficient for comparison. Please include a grader/scoring revision in the fingerprint (or require and update an equivalent immutable field), refresh it during regrade and answer-only scoring, and add a regression proving a regraded report identifies the scorer that actually produced its verdicts.

apurvvkumaria: Refreshing `benchmark_revision` and hashing the verdict code addresses the reported path, but the comparison fingerprint is still incomplete. `bench/report.py::summarize()` calculates the published overall, per-type, difficulty, freshness, and evidence numbers, yet `report.py` is absent from `_SCORER`. A change to its grouping or rounding can therefore change the reported score while leaving the fingerprint identical, even though the submission guide says matching fingerprints are sufficient. Please include every score-producing module, including `report.py`, in the scorer revision and add a regression that proves an aggregation change changes the fingerprint.

examples/tools/agent-memory-benchmark/bench/runner.py
apurvvkumaria: The proxy-silent case now fails closed and is no longer cost-comparable; that resolves the blocking direction of this thread. Non-blocking cleanup: the reverse declaration mismatch is accepted. `_accounting_method("local", 3, 0)` returns `local-unmeasured`, the report stays valid, and its description says no cost was measured even though three requests crossed the proxy. Please reject or accurately report `local` plus observed proxy traffic, with a regression for that mismatch.

apurvvkumaria:
Exact-head follow-up for `0917217f`:

**Blocking:** a missing answer is passed to `grade()` as an empty string, and the abstention grader treats an empty string as a correct refusal. I reproduced this against the shipped corpus: omitting all 13 abstention rows scores all 13 as correct, and 13 missing answers are below the current `> len(questions) // 2` invalidation threshold, so the run remains valid. `tools/regrade.py` has the same behavior and does not record missing answers. Please distinguish an absent row from an explicit abstention, fail incomplete submissions closed, and add runner and regrade regressions showing an omitted abstention answer cannot earn credit.

**PR-description cleanup:** the early claim that the answer key is “out of reach” contradicts the later known gap and the narrowed repository documentation. Please use the narrowed claim consistently. The exact-head license check now reports 560 checked source files, not 550.

The complete 123-test example suite, DCO, SPDX, catalog, and governance checks pass on this head.

examples/tools/agent-memory-benchmark/bench/runner.py
```diff
@@ -0,0 +1,521 @@
...
+        )
+        usage = proxy.usage.snapshot()
+
+    answers = _read_answers(answers_path)
+    verdicts: list[Verdict] = []
+    for question in questions:
+        answered = question["id"] in answers
+        row = answers.get(question["id"], {})
+        verdicts.append(
+            grade(question["id"], str(row.get("answer", "")), row.get("source_ids"),
```
apurvvkumaria: The missing-row path now fails closed, but an incomplete row still earns abstention credit. `answered` checks only whether the ID exists, so both `{"id": "st-abstain"}` and `{"id": "st-abstain", "answer": ""}` reach the grader as an answered empty string; `_is_abstention()` then scores both correct. I reproduced both on this head. The documented contract requires the system to say plainly that the corpus does not support an answer. Please require a present, non-empty answer (or score an empty answer as incorrect) and add runner and regrade regressions for ID-only and empty-answer rows.

examples/tools/agent-memory-benchmark/tools/regrade.py
```diff
@@ -0,0 +1,117 @@
...
+    # The verdicts in this report were produced by the benchmark running now,
+    # not by whatever produced the original, so the revision must move with
+    # them. Leaving the old one made the report name a scorer that did not
+    # score it.
+    report["benchmark_revision"] = _git_revision(REPO)
+    report["answers_missing"] = [q["id"] for q in questions if q["id"] not in answers]
+    if report["answers_missing"]:
```
apurvvkumaria, line 109: Regrading does not clear stale missing-answer invalidation. If the first pass writes `valid: false` and `invalid_reason` because one answer is absent, then the contributor adds that answer and runs `regrade.py` again, `answers_missing` becomes empty but the old invalid fields remain. The resulting report claims it is invalid for an omission it no longer has. Please recompute this validation state on every regrade, while preserving independent accounting failures, and add an incomplete -> corrected -> regraded regression.

examples/tools/agent-memory-benchmark/tests/test_missing_answers.py
```diff
@@ -0,0 +1,126 @@
...
+def test_an_aggregation_change_changes_the_fingerprint(tmp_path):
+    """report.py produces the published numbers, so it is part of the scorer."""
+    from bench.fingerprint import _SCORER, scorer_revision
+
+    assert "report.py" in _SCORER, "the aggregation module must be in the scorer revision"
+    before = scorer_revision()
+    target = REPO / "bench" / "report.py"
+    original = target.read_bytes()
+    try:
+        target.write_bytes(original + b"\n# an aggregation change\n")
```
apurvvkumaria: Non-blocking: this regression temporarily edits the checked-in `bench/report.py`. An interrupted or parallel test run can leave the source tree dirty, which conflicts with the repository requirement that checks be teardown-safe. Please exercise the hash against an isolated temporary module tree or make the scorer helper accept a source directory for testing.

apurvvkumaria (approved):
Approved at 6a0ae141605773f86ded9b7a1a13847359f8bb09. Missing, blank, and non-string answers now fail closed; regrading remains compatible with stored reports that use the legacy accounting shape; and the focused regressions pass with the full 135-test suite. All required repository checks are green.

PR #142, opened 2026-08-26
feat(memory-driven-chief-of-staff): read the user's mailbox into the intake

rebelle5868:
## Related Issue

Related: #122

## Description

The third connector phase, and the one that can keep the stronger deletion guarantee. Slack offers no way to learn that a message was removed — its absence from a bounded read cannot be told apart from it lying outside the window — so Slack content ages out on the retention pass. Microsoft Graph delta reports that a message left the tracked folder. The collector then looks up its immutable `internetMessageId`: a match in another folder is a move, while no match is treated as a deletion.

**Deletion is reconciled rather than waited out.** A message deleted at the source is tombstoned on the next synchronisation and its body cleared at once. The row stays: obligations and events hang off `source_id`, and removing it would break the record of why something was ranked. `deleted_at` and `body_cleared_at` are separate columns because one is the person deleting and the other is this recipe ageing text out on its own schedule. The removal-tracking columns, `deleted_at` and `internet_message_id`, are schema v4, migrated in place and tested against `schema-v3.sql`, the frozen schema shipped in #140.

**The delta cursor is the watermark, which removes a class of bug rather than fixing one.** A cursor is opaque and only issued when a round completes, so it cannot be constructed from a message this run happened to see. "Only a complete crawl may advance the watermark" is enforced by the protocol rather than by remembering to write it correctly — the rule the Slack collector needed three rounds to get right. An interrupted round saves its page link and resumes; an expired cursor is a fresh round rather than a failed run.

**The first synchronisation is bounded to a window the operator chooses.** Seven days by default, `GRAPH_BACKFILL_DAYS` for 1 to 3650, through the same validator the other bounded settings use. I checked against the live service that the delta endpoint accepts a `receivedDateTime` filter on the initial request; without that the choice would have been between synchronising a whole mailbox and not learning about deletions at all. The window decides where the round starts and nothing else — the cursor carries no filter, so once the baseline exists every later change is reported, including an older message being deleted.

**Two shared pieces extracted rather than copied.** The storage-encryption gate and the field-by-field profile validation now live in `require-encrypted-storage.sh` and `validate-provider-profile.sh`, sourced by both setup flows. A second copy of a security gate is a copy that drifts. The validator is parameterised by host and credential key; `setup-slack.sh` passes `slack.com` and behaves as before.

Extracting the gate broke a test in a way worth naming: `test_an_unconfirmed_prerequisite_aborts` searched `setup-slack.sh` between two strings for `exit 1`, the window emptied when the gate moved, and it passed on nothing — the same shape that let the reuse-path bypass ship. It runs the script now.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — 569 files
- [x] `git diff --check`
- [x] Recipe suite: 564 tests across thirteen files, `OK`
- [x] Against a real mailbox on Linux under Hermes 0.19.0: a seven-day first round synchronised 50 messages across 6 pages and stored a cursor; the next tick added nothing and stayed synchronised; moving a message to Deleted Items and re-running produced `removed: 1` with the row and metadata intact, the body gone, and both timestamps set
- [x] Confirmed against the live service that `messages/delta` accepts a `receivedDateTime` filter on the initial request and that the resulting cursor carries none
- [x] Each behaviour checked by reverting it and confirming the assertions turn red — move-is-not-deletion, immediate body clearing, partial crawls not advancing, resumption, the window bound, expired-cursor recovery

- Result: `docs-updated`
- Evidence or justification: `docs/set-up-graph.md` — new, covering the application registration, device-code sign-in, window choice and revocation; `README.md` — both collectors in the file map and the test table, and the Graph deletion sentence moved from future to present tense now that something implements it.
- Reviewer: @rebelle5868

## Release And Compliance

### DIFF
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
@@ -41,6 +41,7 @@ Exit codes are the contract `select_intake.py` reads:
 from __future__ import annotations
 
 import json
+import sqlite3
 import os
 import sys
 import time
@@ -86,10 +87,6 @@ MAX_BACKFILL_DAYS = 3650
 # and the tick after that one still has to finish.
 REQUEST_BUDGET = 10
 
-# How many message identities are remembered for answering "moved or deleted".
-# A removal concerns something collected recently; an unbounded map would grow
-# with the mailbox for no benefit.
-REMEMBERED_IDENTITIES = 5000
 
 # How long one run may spend being told to wait.
 #
@@ -319,8 +316,13 @@ def first_round_url(days: int) -> str:
     return "/me/mailFolders/inbox/messages/delta?" + query
 
 
-def still_in_mailbox(source_id: str, token: str, budget: Budget) -> bool:
-    """Is this message somewhere else in the mailbox, or actually gone?
+def still_in_mailbox(source_id: str, token: str, budget: Budget) -> str:
+    """Is this message somewhere else in the mailbox, gone, or unanswerable?
+
+    Three answers, not two. A bool had to spend "I could not find out" on one
+    of the other two, and both were wrong: as "gone" it cleared the body of a
+    message that had merely been filed away, and as "present" it let a real
+    deletion pass unrecorded.
 
     Asked by `internetMessageId`, which a message keeps when it moves; the
     per-folder id does not, so it cannot be used here.
@@ -336,13 +338,18 @@ def still_in_mailbox(source_id: str, token: str, budget: Budget) -> bool:
     Stopping costs a re-read of pages this tick already handled, which is
     idempotent, and the next tick asks the question again.
     """
-    known = read_state().get("message_ids", {})
-    internet_id = known.get(source_id)
+    with sqlite3.connect(ledger_path()) as conn:
+        row = conn.execute(
+            "SELECT internet_message_id FROM items WHERE source_id = ?",
+            (source_id,)).fetchone()
+    internet_id = row[0] if row else None
     if not internet_id:
-        # Never seen with an identity of its own — collected before this
-        # existed, or the field was absent. Treat a removal as a removal
-        # rather than guessing that it moved.
-        return False
+        # Nothing to ask about: a row collected before this column existed,
+        # or a message that arrived without the field. Unknown, and unknown
+        # is not "deleted" — reading it that way tombstoned the row and
+        # cleared its body, which is the irreversible direction and was
+        # reached without contacting Graph at all.
+        return "unknown"
     query = urllib.parse.quote(
         "internetMessageId eq '%s'" % internet_id.replace("'", "''"), safe="")
     try:
@@ -355,7 +362,7 @@ def still_in_mailbox(source_id: str, token: str, budget: Budget) -> bool:
             "could not establish whether a removed message was deleted or "
             f"moved ({exc}); the round stops here rather than advancing past "
             "a removal it did not resolve", exc.kind) from exc
-    return bool(found.get("value"))
+    return "present" if found.get("value") else "gone"
 
 
 def tombstone(conn, source_ids: list[str]) -> int:
@@ -386,19 +393,33 @@ def tombstone(conn, source_ids: list[str]) -> int:
     return cursor.rowcount
 
 
-def commit(items: list[dict[str, Any]], removed: list[str]) -> tuple[int, int]:
-    """Rows and tombstones, in one transaction, taken after the fetch.
+def commit_rows(items: list[dict[str, Any]]) -> int:
+    """The page's messages, in their own transaction.
 
-    Holding a write transaction open across a network call holds a write lock
-    for the length of that call, and the user's own corrections queue behind
-    it. The delta cursor is saved by the caller and only after this returns:
-    a cursor recorded over rows that were never written would have the next
+    Separate from the tombstones, and before them, for two reasons. A write
+    transaction held across a network call holds a write lock for the length
+    of that call, and the user's own corrections queue behind it. And the
+    question a removal asks — what is this message's own identity — is
+    answered from the row, so the rows have to be there before it is asked;
+    a message collected in one page and removed in a later page of the same
+    round is otherwise unanswerable.
+
+    The delta cursor is saved by the caller and only after this returns: a
+    cursor recorded over rows that were never written would have the next
     round start past them.
     """
+    if not items:
+        return 0
     with write_txn() as conn:
-        added = insert_items(conn, items) if items else 0
-        gone = tombstone(conn, removed)
-    return added, gone
+        return insert_items(conn, items)
+
+
+def commit_tombstones(removed: list[str]) -> int:
+    """What the source no longer has, once that has actually been asked."""
+    if not removed:
+        return 0
+    with write_txn() as conn:
+        return tombstone(conn, removed)
 
 
 def collect(token: str, address: str, state: dict[str, Any],
@@ -424,7 +445,7 @@ def collect(token: str, address: str, state: dict[str, Any],
     else:
         url, absolute = first_round_url(days), False
 
-    added_total = removed_total = moved = pages = 0
+    added_total = removed_total = moved = unresolved = pages = 0
     delta_link = None
 
     while url and pages < REQUEST_BUDGET:
@@ -449,30 +470,28 @@ def collect(token: str, address: str, state: dict[str, Any],
         # own identity rather than its position. If a search of the mailbox
         # still finds it, it moved; if it does not, it is gone. One request
         # per removal, and removals are rare beside messages.
+        #
+        # That identity is kept on the row, by `normalize`, rather than in a
+        # map beside the cursor. The map was bounded at five thousand and
+        # evicted oldest-first, so filing away a message older than that
+        # produced a removal nobody could ask about — and the code read the
+        # absence as a deletion, cleared the body, and never made a request.
+        # The row is where the message already is, and it does not evict.
         gone_ids = [m["id"] for m in batch
                     if isinstance(m.get("@removed"), dict) and m.get("id")]
         present = [m for m in batch if "@removed" not in m and m.get("id")]
 
-        removals = [source_id for source_id in gone_ids
-                    if not still_in_mailbox(source_id, token, budget)]
-        moved += len(gone_ids) - len(removals)
-
-        # Remember each message's own identity, so a later removal can be
-        # asked about. Bounded, and oldest first out: a removal concerns
-        # something collected recently, and an unbounded map would grow with
-        # the mailbox.
-        known = state.setdefault("message_ids", {})
-        for m in present:
-            if m.get("internetMessageId"):
-                known[m["id"]] = m["internetMessageId"]
-        if len(known) > REMEMBERED_IDENTITIES:
-            for stale in list(known)[:len(known) - REMEMBERED_IDENTITIES]:
-                known.pop(stale, None)
-
+        # Written before the removals are resolved, so a message collected
+        # earlier in this same round can be asked about in a later page.
         items = [graph_message_to_item(m, address) for m in present]
-        added, gone = commit(items, removals)
-        added_total += added
-        removed_total += gone
+        added_total += commit_rows(items)
+
+        verdicts = {source_id: still_in_mailbox(source_id, token, budget)
+                    for source_id in gone_ids}
+        moved += sum(1 for v in verdicts.values() if v == "present")
+        unresolved += sum(1 for v in verdicts.values() if v == "unknown")
+        removed_total += commit_tombstones(
+            [s for s, verdict in verdicts.items() if verdict == "gone"])
 
         delta_link = payload.get("@odata.deltaLink")
         url = payload.get("@odata.nextLink")
@@ -494,7 +513,15 @@ def collect(token: str, address: str, state: dict[str, Any],
         complete = False
 
     return {"source": "email", "scope": "inbox", "added": added_total,
-            "removed": removed_total, "moved": moved, "pages": pages,
+            "removed": removed_total, "moved": moved,
+            # Removals that left the folder while the row had no identity of
+            # its own to ask about — collected before the column existed, or
+            # arrived without the field. Left alone rather than tombstoned,
+            # and counted rather than passed over in silence, because the
+            # alternative was clearing the body of a message that had only
+            # been filed away.
+            "unresolved_removals": unresolved,
+            "pages": pages,
             "resumed": resuming, "complete": complete,
             "synchronised": bool(state.get("delta"))}
 
@@ -550,7 +577,7 @@ def main(argv: list[str] | None = None) -> int:
             # would synchronise one mailbox against another's position.
             print("the mailbox has changed; discarding the previous "
                   "synchronisation state", file=sys.stderr)
-            for key in ("delta", "next", "message_ids"):
+            for key in ("delta", "next"):
                 state.pop(key, None)
         elif refresh:
             state.pop("delta", None)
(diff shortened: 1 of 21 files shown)


examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-graph.sh
```diff
@@ -0,0 +1,305 @@
...
+SANDBOX="${OPENSHELL_SANDBOX_NAME:-${SANDBOX_NAME:-hermes}}"
+USABLE_KEY="MS_GRAPH_ACCESS_TOKEN"
+WANT_HOST="graph.microsoft.com"
+
+# Microsoft's own public client id for the device-code flow, and the tenant
+# most people want. Both are overridable: an organisation that registers its
+# own application passes its ids rather than editing this file.
+CLIENT_ID="${GRAPH_CLIENT_ID:-}"
+TENANT_ID="${GRAPH_TENANT_ID:-common}"
+SCOPES="offline_access https://graph.microsoft.com/Mail.Read"
```
apurvvkumaria: `identity()` calls `GET /me`, but Microsoft documents `User.Read` as the least-privileged delegated permission for that endpoint. This flow requests only `Mail.Read`, so the documented setup depends on an additional permission that it neither requests nor names. Please either remove the `/me` dependency or explicitly request and document `User.Read`, with a regression that exercises the declared scope set. Reference: https://learn.microsoft.com/en-us/graph/api/user-get?view=graph-rest-1.0

examples/recipes/nvidia/memory-driven-chief-of-staff/providers/graph-user.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - name: graph_access_token
+    description: Microsoft Graph delegated access token
+    env_vars: [MS_GRAPH_ACCESS_TOKEN]
+    required: true
+    auth_style: bearer
+    header_name: authorization
+    refresh:
+      strategy: oauth2_refresh_token
+      token_url: https://login.microsoftonline.com/common/oauth2/v2.0/token
+      scopes: [https://graph.microsoft.com/.default]
```
apurvvkumaria: Using `.default` during refresh contradicts the stated minimum-permission boundary. Microsoft documents that `.default` returns the Graph permissions already granted for that client and user, which can include more than this recipe needs. Please refresh with the exact required scopes, or otherwise enforce and document the effective scope set. Reference: https://learn.microsoft.com/en-us/entra/identity-platform/scopes-oidc#the-default-scope

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
```diff
@@ -0,0 +1,447 @@
...
+        pages += 1
+
+        batch = payload.get("value") or []
+        # A delta page mixes changes and removals. `@removed` carries a reason
+        # — `deleted` or `changed` — and only the first is a deletion; the
+        # second means the item moved out of the scope being synchronised,
+        # which is not the same thing and must not tombstone it.
+        removals = [m["id"] for m in batch
+                    if isinstance(m.get("@removed"), dict)
+                    and m["@removed"].get("reason") == "deleted"
```
apurvvkumaria: This does not distinguish deletion from a move out of the Inbox. Microsoft documents that message delta returns `@removed.reason == "deleted"` when an item is deleted **or moved from the tracked folder**. The current test invents `reason == "changed"` for a move, so filing a message into another folder can be recorded as a user deletion and clear its body. Please align the implementation, documentation, and regression with Graph's actual move response. Reference: https://learn.microsoft.com/en-us/graph/api/message-delta?view=graph-rest-1.0

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
```diff
@@ -0,0 +1,447 @@
...
+                return json.loads(response.read().decode("utf-8"))
+        except urllib.error.HTTPError as exc:
+            if exc.code in (401, 403):
+                raise GraphError(
+                    "Microsoft Graph refused the credential (HTTP %d)." %
+                    exc.code,
+                    "credential") from exc
+            if exc.code == 410:
+                raise GraphError("delta cursor expired", "resync") from exc
+            if exc.code == 429:
```
apurvvkumaria, line 254: The retry loop has no attempt or elapsed-time bound when Graph repeatedly returns `429` with `Retry-After` at or below 30 seconds. The page budget counts only successful responses, and the scheduled caller supplies no subprocess timeout, so one tick can hang indefinitely. I reproduced 9,248 retries without termination. Please add a total retry/time bound that returns the documented rate-limit exit code, plus a repeated-429 regression; the current test named for rate limiting sends HTTP 500 instead.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-graph.sh
```diff
@@ -0,0 +1,305 @@
...
+    fi
+    reusable="$name"
+    break
+  fi
+done < <(printf '%s\n' "$attached" \
+         | sed 's/\x1b\[[0-9;]*m//g' \
+         | awk 'NR > 1 && NF { print $1 }')
+
+if [[ -n "$reusable" ]]; then
+  echo "     reusing attached provider '$reusable' (type and policy verified)"
```
apurvvkumaria, line 127: The reuse path verifies the provider type and endpoint profile but never verifies that refresh is configured or healthy. A provider that exposes the key but has no refresh chain can therefore print `Nothing to do` and exit successfully. Please check `openshell provider refresh status` and the expected refresh strategy before declaring the provider reusable, with a setup regression for missing or mismatched refresh state.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
```diff
@@ -0,0 +1,447 @@
...
+             refresh: bool = False) -> dict[str, Any]:
+    """Whose mailbox this is, cached.
+
+    The address decides addressing — being a To recipient is being asked,
+    being copied is being informed — so it is needed on every message and
+    fetched once. Keyed on the credential, so replacing the token re-probes
+    rather than inheriting the previous mailbox's identity along with its
+    delta cursor.
+    """
+    mark = fingerprint(token)
```
apurvvkumaria: Inside the sandbox, `token` is the stable OpenShell placeholder, not the rotating or replaced bearer token. Hashing it therefore does not re-probe identity after `FORCE_REAUTH`, despite the docstring's claim, and the previous mailbox identity and delta cursor can remain attached to a different account. Please re-check a durable Graph user identifier when credentials are replaced and clear `next`/`delta` when the mailbox changes, with a regression covering reauthorization to another account.

apurvvkumaria (changes requested):
The full 456-test recipe suite and all required repository checks pass, but the mailbox connector is not merge-ready yet. The inline findings cover the Graph deletion contract, OAuth scope boundary, bounded rate-limit handling, provider reuse, and mailbox identity state. Please also update the main README: it still says the Graph collector is absent, describes only Slack as networked, and documents only the Slack egress and credential boundary. Those statements now contradict the shipped collector and affect security, privacy, and setup expectations.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-graph.sh
rebelle5868: Correct, and it worked only for whoever already had the permission. Both `Mail.Read` and `User.Read` are requested now, and the reason is stated where the scopes are set: `/me` is how the collector learns whose mailbox it is reading, which decides whether a message addressed the user or merely copied them.

The first version of the test for this searched the whole setup script for `User.Read`, which also appears in the comment explaining why it is needed — so deleting the scope left it green. It reads the `SCOPES=` line now.

examples/recipes/nvidia/memory-driven-chief-of-staff/providers/graph-user.yaml
rebelle5868: Agreed. `.default` returns whatever the client has already been granted, which is more than this needs and grows silently as an administrator grants more — the opposite of the boundary the profile declares two lines further down. The three scopes are named.

The same loose-assertion problem applied here: the test looked for `.default` anywhere in the file, and the file names it in the comment explaining why it is not used. It reads the scope list now.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
rebelle5868: Reproduced, and worse than a missing case: I invented the contract and then tested my invention. There is no documented `changed` reason and the service does not send one. Measured against the live API — moving a message to Archive produces exactly the removal a deletion produces, and the per-folder id changes, so the id cannot answer it either. Archiving was therefore recorded as the person deleting, and the body cleared at once.

What survives a move is `internetMessageId`, which belongs to the message rather than to its position. A removal now asks whether the mailbox still holds it: found means filed, not found means gone. One request per removal, and removals are rare beside messages; the identities are remembered bounded and oldest-out, since a removal concerns something collected recently. `moved` and `removed` are reported separately.

A test asserts the lookup is actually made. Without it the two cases are indistinguishable, so a test that never sees the request would not be testing the distinction.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
rebelle5868: Reproduced. Each individual wait was bounded and the sum was not, so repeated `429`s inside the per-wait limit looped with no exit — the page budget counts successful responses and a cron pre-step has no timeout of its own.

There is a total bound now, and exceeding it reports rate limiting so the next tick continues from where this one stopped rather than the tick hanging.

You were also right that the test named for rate limiting was sending a 500. It sends relentless 429s now, and removing the bound makes it hang rather than fail — which is the honest failure mode for this defect, and the reason the original was invisible.

examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/setup-graph.sh
rebelle5868: Reproduced. A provider with the right key, the right type and the right endpoint policy but no refresh chain printed "Nothing to do" and exited zero, then stopped within the hour — the least useful moment to be told nothing.

It checks `provider refresh status` and the strategy now, as the Slack setup does, with regressions for a missing refresh chain and for the wrong strategy, both driving the script against a stubbed `openshell`.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
rebelle5868: This is the deepest of the six and you are right about the mechanism. The digest was of a value the gateway substitutes, so it stayed constant across a re-authorisation to a different account — and the delta cursor, the resume link and the remembered message identities would all have been carried over to somebody else's mailbox.

`/me` is asked every run now and the mailbox address compared. It is durable, it belongs to the account rather than to the credential presented for it, and it is the addressing input every message needs anyway, so the cost is a request that was already being made on the first tick. When the address changes, `delta`, `next` and `message_ids` are discarded together.

The first test for this asserted that the recorded address had changed, which proves nothing — it is rewritten on every run regardless. It asserts that the old cursor is not followed and that a fresh round starts.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
apurvvkumaria: One blocker remains. `MAX_TOTAL_BACKOFF_SECONDS` is initialized inside `call()`, so the budget resets for `/me`, every delta page, and every removal lookup rather than bounding the scheduled tick. More importantly, `still_in_mailbox()` catches every `GraphError` and returns `true`; if that lookup receives a 429, credential error, or service failure, the collector treats the removal as a move and can still advance the delta cursor, so the deletion may never be retried. Please carry one tick-wide deadline or retry budget through all Graph calls and propagate removal-lookup failures so the round does not advance. Add regressions for repeated 429s across consecutive requests and during a removal lookup.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/ingest_graph.py
```diff
@@ -0,0 +1,591 @@
...
+    the cursor advances past the page that carried it, and nothing ever asks
+    again: the message stays in the store as though it were never deleted,
+    permanently, and the person who deleted it is never told otherwise.
+
+    Stopping costs a re-read of pages this tick already handled, which is
+    idempotent, and the next tick asks the question again.
+    """
+    known = read_state().get("message_ids", {})
+    internet_id = known.get(source_id)
+    if not internet_id:
```
apurvvkumaria, line 346: Blocker: `still_in_mailbox()` returns `false` without querying Graph whenever `message_ids` has no entry for the removed source ID. The map is capped at 5,000 and evicts older entries, so a user can later move an older Inbox message whose identity was evicted; this path then treats the move as a deletion, tombstones the row, and clears its body. I reproduced the branch with an empty map: it returned `false` and the mailbox lookup was never called. Please fail closed when the identity is unavailable, or persist the `internetMessageId` for every tracked message and use the in-memory state for this tick. Add regressions for an evicted or legacy identity and for an identity learned earlier in the same synchronization round.

rebelle5868: Reproduced, and it is the worst of the three: the destructive branch was the one that made no request at all. An absent identity returned "gone", so the row was tombstoned and its body cleared on the strength of a lookup that never happened — and with the map capped at five thousand and evicting oldest-first, filing away an older message reached exactly that path.

The identity now lives on the row, written by `normalize` from the field the delta query was already requesting. A row does not evict, so the cap, the map and its discard-on-mailbox-change are all gone.

I took the answer to three values rather than fail closed on a bool, because a bool has to spend "I could not find out" on one of the other two and both are wrong: as "gone" it clears a body that should have been kept, and as "present" it lets a real deletion pass unrecorded and, since a removal is reported once, permanently. `unknown` tombstones nothing and is reported as `unresolved_removals` rather than passed over in silence.

Your second point is handled by writing the page's rows before resolving that page's removals, so a message collected on one page and removed on a later page of the same round is answerable immediately rather than at the next tick. That splits the write into two transactions, which is also what the existing comment about not holding a write lock across a network call was asking for.

Regressions: an evicted or legacy identity, asserting both that no request was made and that the body survived; an identity learned earlier in the same round; the same arrangement where the message merely moved; and another mailbox's removal naming an id this store has never seen. Reverting `unknown` back to `gone` turns two of them red.

Still outstanding and unchanged: the v3 to v4 renumber waits on #140 landing, since #142 alone would have a hole at v3. `internet_message_id` ships in the same migration as `deleted_at` because they are one feature, so both columns move to v4 together.

apurvvkumaria:
Remaining blocker now that #140 is merged: please rebase onto current `main`, preserve #140 stable source identity as schema v3, and move `_add_removal_tracking` (`deleted_at` and `internet_message_id`) to schema v4. Update `SCHEMA_VERSION`, the `schema.sql` meta version, migration fixtures and expected migration sequences, then run the complete recipe suite. The current head still assigns removal tracking to v3, which now collides with the merged v3 migration.

apurvvkumaria (approved):
Approved after the Graph removal, authorization, retry-budget, mailbox-identity, and schema-migration issues were addressed. The v3-to-v4 migration is now exercised against the exact schema shipped by #140, the full 564-test recipe suite passes, and the required repository checks are green.

PR #143, opened 2026-08-26
feat(tools): reject three grading attacks, and add reference results for corpus A

rebelle5868:
## Related Issue

Follows #136 and #141, where reference results were deferred to a separate pull request.

## Description

Two changes, in two commits, in that order.

### 1. A grading contract that did not survive an adversary

Two independent reviews of the upstream benchmark found three attacks that scored full marks against every shipped key. Two of them reproduce against the version merged in #141, and I reproduced them before fixing anything:

```
refuse, then fabricate      13/13 corpus A abstentions, 12/12 corpus B
deny every accepted value   blocked already, by luck rather than by design
```

`"No evidence in the corpus. The cutover completed"` satisfied the abstention check before the reject check ever ran, so a system could take every abstention question by declining and then asserting the fabrication anyway. `"None of these are correct: 50%; 75%"` lists both keyed answers in order to reject them, and a containment test matches them regardless.

The fix is a negation scope, ported from upstream: a value counts as asserted only when it does not sit inside a denial, scoped back to the sentence break. Two refinements each cost a round of debugging there — a denial introducing a list governs the whole list, and a contrast ends the scope so a correct correction is not read as denying its own answer. A sentence break requires whitespace after the dot, or a path like `src/atlas/auth/quillon_callback.py` splits mid-token.

`tests/test_grader_mutations.py` generates the attacks from the shipped keys themselves, so a key added later is attacked without anyone remembering to attack it. All three now score zero on both corpora.

Two contract differences with upstream are resolved in favour of the stricter reading already merged here: an empty answer fails everywhere rather than counting as a refusal, and `grade()` keeps the `answered=` parameter that distinguishes silence from an abstention.

**This changes every fingerprint**, because the scorer hash covers `grader.py`. The published content hashes are regenerated in the same commit. That is also why the results are in this pull request rather than a later one: a result produced before this fix carries a scorer hash that no longer describes anything, and `docs/SUBMITTING.md` tells readers that matching fingerprints are what makes two rows comparable.

### 2. Reference results for corpus A

One pair of runs, both on NVIDIA Nemotron 3 Ultra, both graded by the grader in this repository:

| Metric | Agentic RAG | Self-model | Difference |
| --- | ---: | ---: | ---: |
| Overall accuracy (186 questions) | 82.8% | 89.8% | +7.0 pp |
| Hard questions (31) | 67.7% | 87.1% | +19.4 pp |
| Tracking facts that changed over time (5) | 60.0% | 100.0% | +40.0 pp |
| Point-in-time reasoning (6) | 33.3% | 66.7% | +33.3 pp |
| Entity disambiguation (15) | 66.7% | 86.7% | +20.0 pp |
| Multi-source synthesis (73) | 87.7% | 94.5% | +6.8 pp |
| Refusing to answer when the corpus cannot (13) | 100.0% | 76.9% | -23.1 pp |
| Single-hop lookup (30) | 86.7% | 83.3% | -3.3 pp |
| Citation coverage (186) | 92.5% | 97.8% | +5.4 pp |

Cost is reported separately, never folded into the above, and never as a ratio: each run's counts are what that run observed, and with no forwarded-call record in either report nothing establishes that the two counted the same events, so both set `comparable_on_cost` to false. The agentic baseline used `adapters/agentic_rag/` exactly as it ships, at its default three rounds, and the report records that adapter's file hash so a reader can tell whether it has moved since.

Both runs predate the forwarded-call record the harness now keeps, so neither is marked cost-comparable: the reports publish the calls each run observed, ingest included, and leave forwarded and uncounted as null rather than asserting a total the artifact cannot establish.

**Three limits are stated where the numbers are, in `results/README.md`, not in a footnote.**

- **Corpus A only.** Corpus B ships and was not run on this model.
- **One base model**, where `docs/methodology.md` asks a submission to run at least two and publish the difference. This is a worked example of the output format, not a submission that meets that bar.
- **Not reproducible as-is.** These answers were written against the corpus as it stood before publication, when a project, a class, a mailbox folder, a documentation path, an email domain and four question ids carried different names.

On that last point: the same substitution applied to the corpus at publication was applied to the answers, because an answer is derived from the corpus that produced it. Each run also ships `answers.as-answered.jsonl`, the untransformed output, so the transform can be checked rather than believed. The effect is measured, published, and derived by a test that regrades both files rather than restated by hand — 80.1% → 82.8% and 87.1% → 89.8% — and the full map is in each `report.json` under `provenance_note`.

The id map does more work than the text map, and the reason is worth stating: four question ids were renamed at publication, and without mapping them those four answers look absent. An absent answer invalidates a run, which is the runner behaving correctly, and it is also why a transformed result is weaker evidence than a fresh one.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — all 565 checked source files have SPDX headers
- [x] `python3 scripts/check_catalog_parity.py --check` — 17 entries across 6 categories
- [x] `python3 scripts/check_pr_title.py "<title>"`
- [x] `git diff --check`
- [x] `python3 -m pytest tests/` in the example directory — **219 passed**, offline

New coverage:

| Test file | Tests | Covers |
| --- | ---: | --- |
| `test_grader_mutations.py` | 36 | the three original attacks generated from the shipped keys, plus a thirty-two-case contract covering both directions: an attack that scores, and a valid correction that is rejected |
| `test_published_results.py` | 31 | each run ships its four artifacts plus the untransformed answers; the published map reproduces `answers.jsonl` from `answers.as-answered.jsonl` row for row; nothing changed that the map does not touch; the report follows the current schema; the transform effect is what regrading actually gives; accounting claims only what the artifact establishes; the shipped adapter's hash matches the adapter; no published prose states a cost ratio while a run says its accounting cannot support one; any price a report carries is stated in the prose; the README's rename count matches the map; every question type the self-model loses is reported on the results page; a generated run is ignored rather than merely untracked |
| `test_missing_answers.py` | 11 | silence is not an abstention, in the runner and in regrade |
| `test_accounting.py` | 9 | counted, missing-usage, bypassed, locally-hosted and mis-declared runs |
| `test_no_registrable_domains.py` | 6 | tracked public artifacts use reserved domains and do not expose contributor-machine paths |

- Result: `docs-updated`
- Evidence or justification: `results/README.md` is new and carries the numbers together with all three limits. `README.md` gains a Results section pointing at it, `docs/SUBMITTING.md` notes that reference runs ship, and `.gitignore` stops ignoring `results/runs` wholesale while keeping run state and proxy logs out.
- Reviewer: Xuan Wu
- [x] Changed user-facing text follows the writing guide and controlled-word list
- [x] A public contributor can understand the changed text without internal company context
- [x] I reviewed any agent-generated text before submission

## Release And Compliance

- [x] No secrets or credentials are included
- [x] No nonpublic project names, environment names, hostnames, URLs, ticket identifiers, workspace paths, logs, screenshots, or configuration values are included
- [x] Third-party dependency changes are reflected in `THIRD-PARTY-NOTICES` — no dependency is added
- [x] Public content uses sanitized examples and placeholders instead of private values
- [x] I added my DCO sign-off declaration to this pull request description


examples/tools/agent-memory-benchmark/bench/grader.py
```diff
@@ -82,27 +78,160 @@ def _hit(answer: str, needles: Iterable[str]) -> str | None:
...
+    cutover completed" into one clause and let the leading refusal suppress the
+    fabrication that followed it.
+    """
+    if not contains(answer, value):
+        return False
+    scoped, needle = _scoped(answer), _scoped(value)
+    if not needle or needle not in scoped:
+        # Matched only after full normalisation (a date form, a unicode
+        # variant). No clause to inspect, so trust the match.
+        return True
```
apurvvkumaria, line 316: **Blocker:** This fallback still accepts a denied normalized alias. With `accept: ["2026-07-14"]`, `None of these dates is correct: July 14, 2026` grades correct because `contains()` normalizes the date but the scoped needle does not match literally. I also reproduced `50% is not correct` and `Neither 50%, nor 75% is correct` grading correct, while valid corrections such as `Sofia Ramos rather than Kofi approved it` fail when `Kofi` is rejected. Please preserve match spans through normalization and handle post-value negation, denial lists, and contrast punctuation. Add regression tests for these cases so the adversarial contract covers more than the one generated sentence shape.

examples/tools/agent-memory-benchmark/tests/test_published_results.py
```diff
@@ -0,0 +1,108 @@
...
+    assert fresh["summary"] == _report(run)["summary"], (
+        f"{run.name}: re-scoring the shipped answers does not give the shipped summary"
+    )
+
+
+@pytest.mark.parametrize("run", RUNS, ids=lambda p: p.name)
+def test_every_run_is_valid(run: Path):
+    report = _report(run)
+    assert report.get("valid") is not False, report.get("invalid_reason")
+    assert not report.get("answers_missing")
```
apurvvkumaria, line 84: **Blocker:** These are newly published reference results and are described as worked examples of the output format, but both `report.json` files retain the legacy shape: no `schema_version`, no adapter revision, no trial metadata, and a bare `"accounting": "proxy"` instead of the current structured accounting record. Please migrate both reports to the current schema (including exact model/run parameters and traceable adapter provenance) and make this test enforce that schema rather than accepting legacy output.

examples/tools/agent-memory-benchmark/tests/test_published_results.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+@pytest.mark.parametrize("run", RUNS, ids=lambda p: p.name)
+def test_the_transform_is_disclosed(run: Path):
+    note = _report(run).get("provenance_note")
+    assert note and note.get("answers_transformed") is True
+    assert note.get("substitutions"), "the substitution map must ship with the claim"
```
apurvvkumaria: **Blocker:** This proves only that a non-empty map exists; it does not prove the claim that applying the map to `answers.as-answered.jsonl` produces `answers.jsonl` and changes nothing else. The reports also replace three question-ID mappings with `and three siblings`, so the advertised full map is not present. Please store every text and ID substitution explicitly and add a test that applies that exact map to the as-answered rows and compares the complete transformed rows with `answers.jsonl`.

examples/tools/agent-memory-benchmark/.gitignore
```diff
@@ -3,8 +3,11 @@ __pycache__/
...
 
-# Run artifacts. The harness writes report.json, summary.md, answers.jsonl and
-# verdicts.jsonl under results/runs/<run-id>/ plus whatever memory the system
-# under test built. None of it is source, and the built memory can reach
-# hundreds of megabytes, so a run is kept by whoever produced it.
-results/
+# Run artifacts. A run writes report.json, summary.md, answers.jsonl and
+# verdicts.jsonl, plus whatever memory the system under test built. The memory
+# can reach hundreds of megabytes and is not evidence about a score, so it is
+# never tracked; a run you want to publish is added explicitly.
```
apurvvkumaria, line 9: **Documentation correction:** `README.md` still says ``results/` is not tracked`, which is no longer true now that selected reference runs ship. Please describe the actual rule: generated run state remains ignored, while a deliberately published result is tracked explicitly.

apurvvkumaria (changes requested):
Requesting changes for the three blocking issues in the inline comments: the denial-aware grader is still bypassable and also rejects valid corrections; the answer transformation is not fully specified or mechanically verified; and the newly published reports do not follow the benchmark's current report schema. The README cleanup statement is a smaller documentation correction.

All repository checks pass, the PR is conflict-free, and I reproduced the contributor's 153 passing offline tests. DCO and SPDX are clean. The remaining issues are substantive result-integrity problems that the current tests do not cover.

examples/tools/agent-memory-benchmark/bench/grader.py
apurvvkumaria: Thanks for adding the regression table. The clause splitter still has ordering/nesting failures because it derives one sentence-wide denial flag before splitting and handles a rejecting contrast before an outer denial. I reproduced four remaining cases: `Not Kofi but Sofia approved it` is rejected; `Sofia, not Kofi, approved it` is rejected; `75% is correct; 50% is incorrect` is rejected; and `None of these are correct: Sofia rather than Kofi` is accepted. The first three are valid corrections and the fourth is the same denial attack with a nested contrast. Please address these shapes and add them to the contract table; this blocker remains open.

examples/tools/agent-memory-benchmark/tests/test_published_results.py
apurvvkumaria: The report shape is now current, but the metadata is not yet traceable or internally consistent. `adapter.revision` is only a note pointing to `provenance_note`, which contains no immutable adapter source or revision; for the shipped agentic adapter, record the actual revision or file hash. The self-model `run_parameters` also repeats the agentic-adapter note instead of describing that run. Accounting excludes ingest calls: `usage_raw.calls` totals 1,056 for agentic and 3,278 for self-model, while `forwarded_calls` and `proxy_observed_calls` say 1,048 and 1,675. Because the legacy usage has no forwarded-call record, `uncounted_calls: 0` and `comparable_on_cost: true` are not established by the artifact. Please supply traceable evidence or mark these counts unknown and cost comparison unsupported. Finally, `benchmark_revision` identifies the previous commit as dirty, not a clean revision containing this scorer. This blocker is only partially addressed.

apurvvkumaria (commented):
Follow-up review at the updated head. The complete substitution map and README correction are resolved. The existing change request remains because the grader and report-integrity blockers are still open. Non-blocking: please also update the PR description from 153 tests to the current 172 and refresh its new-coverage counts.

examples/tools/agent-memory-benchmark/results/runs/agentic-rag_corpus-a_nemotron/report.json
```diff
@@ -0,0 +1,195 @@
...
+      "question_ids": {
+        "fresh-helix-launch-date": "fresh-quillon-launch-date",
+        "abstain-helix-launched": "abstain-quillon-launched",
+        "disambig-helix-single-date": "disambig-quillon-single-date",
+        "chain-helix-date": "chain-quillon-date"
+      }
+    },
+    "not_a_rerun": "This is not a run against the published corpus. The systems never saw it. Reproducing these numbers requires re-running the answer phase.",
+    "effect": {
+      "accuracy_as_answered": 0.8172,
```
apurvvkumaria: **Blocker:** These `effect` values still describe the previous grader. Regrading the shipped raw and transformed answers at this head produces `0.8011 -> 0.8280` for agentic and `0.8817 -> 0.9086` for self-model, matching the updated rounded values in `results/README.md`; the reports still say `0.8172 -> 0.8441` and `0.8978 -> 0.9247`. Please update both reports and extend `test_published_results.py` to regrade `answers.as-answered.jsonl` as well as `answers.jsonl`, then assert both computed values against `provenance_note.effect`.

rebelle5868:
All three blockers are addressed at `c943e5f`.

**The clause splitter.** All four shapes reproduced, and your diagnosis was the cause: a sentence-wide denial flag computed before splitting, and a rejecting contrast applied ahead of any enclosing denial. Splitting now happens first and each piece is judged on its own, with denial inherited forward and never backward. A rejecting contrast denies its tail; a resetting marker clears the denial except before a list continuation, where it carries into the next item, and except before another denial, where clearing would hand the answer the value that tail rejects. A denial before a colon governs the list it introduces, and nothing inside can lift itself out.

Fixing that surfaced a defect the previous version had as well: a comma between digits was treated as a clause boundary, so `55,000` was split in half and the value vanished. The oracle fixture caught it — which is the argument for keeping a fixture whose score is known in advance. The contract table is eighteen cases now, both directions, since the first version of this scope failed both.

**Accounting.** You are right that the artifact does not establish those numbers, and I should not have written them. Both reports now publish the calls each run actually recorded, ingest included — 1,056 and 3,278 — with `forwarded_calls` and `uncounted_calls` null and `comparable_on_cost` false, plus a test asserting exactly that shape so a later edit cannot quietly assert a total again. `run_parameters` describes its own run. `benchmark_revision` no longer names a commit it cannot vouch for; the scorer that graded these is identified by `fingerprint.scorer`.

On the adapter revision: the agentic baseline ran `adapters/agentic_rag`, which ships here, so I hashed the adapter tree and the shared lib it imports into the report, with a test asserting the recorded hash matches what ships. If the adapter moves, the published run stops describing it and the test fails. The self-model has no adapter here and keeps a null hash with a note saying so.

**The effect values.** They described the previous grader because the grader moved and the report did not. They are derived now: a test regrades both `answers.as-answered.jsonl` and `answers.jsonl` and compares each against `provenance_note.effect`, so they cannot drift again. Current values are 80.1% → 82.8% and 86.6% → 89.2%; they differ slightly from the ones you computed because the splitter changed again after your comment.

**Scores moved with the grader**, twice, and the direction is worth noting: 84.4% / 92.5% at +8.1 pp became 82.8% / 89.2% at +6.5 pp. The correction lands mostly on the self-model, which phrases more of its answers as a denial followed by a correction — the shape the scope now reads properly rather than crediting.

The PR description is updated to 183 tests with refreshed coverage counts.

examples/tools/agent-memory-benchmark/bench/grader.py
apurvvkumaria: The latest head still has denial shapes outside the regression table. I reproduced valid corrections `The answer is not 50%: it is 75%` and `Sofia rather than Kofi: she approved it` grading false, while denials `50% is wrong` and `Anything except 50%` grade true. Please cover these in both directions and adjust the parser so a colon does not make an earlier denial govern the corrective clause, while post-value and `except` denials remain denied.

examples/tools/agent-memory-benchmark/tests/test_published_results.py
apurvvkumaria: The report metadata now correctly marks these runs as not comparable on cost, but `results/README.md` still says the self-model pays about 1,076 times more to build memory and 17 times more per answer. Those ratios are a cost comparison derived from accounting the artifacts explicitly say is incomplete. Please remove the ratios, report only the independently observed counts, and add a regression that prevents comparative cost claims when `comparable_on_cost` is false.

examples/tools/agent-memory-benchmark/bench/grader.py
```diff
@@ -82,27 +78,258 @@ def _hit(answer: str, needles: Iterable[str]) -> str | None:
...
+# completed") satisfies the abstention check before the reject check ever runs.
+DENIAL_CUES = (
+    "none of these", "none of the above", "neither", "not correct", "incorrect",
+    "is not", "are not", "was not", "were not", "isn't", "aren't", "wasn't",
+    "weren't", "no longer", "rather than", "instead of", "not ", "never ",
+    "cannot", "can't", "don't", "doesn't", "didn't", "no evidence", "not the",
+    # A denial can follow the value instead of preceding it. Anchoring these on
+    # the copula keeps "is wrong" from matching a noun like "the wrong-way
+    # valve", which asserts nothing.
+    "is wrong", "are wrong", "was wrong", "were wrong", "is false", "are false",
```
apurvvkumaria, line 146: Blocker: These cues are still matched as raw substrings, so they are not actually anchored to the copula as the comment says. At this head, `This wrong estimate was corrected to 75%` with `accept: ["75%"]` grades false because `"is wrong"` matches across the end of `This` and the start of `wrong`. The same happens with `This wrong value was replaced with 75%`. Please match these cues with token or regex boundaries and add a regression proving that a valid correction containing `This wrong ... 75%` still asserts `75%`.

examples/tools/agent-memory-benchmark/README.md
```diff
@@ -296,10 +296,23 @@ bench/       runner, grader, accounting proxy, price table
...
 tools/       re-grade stored answers against the current rules
 tests/       the benchmark's own tests
 ```
 
+## Results
+
+One pair of runs ships as a worked example: a self-model and an agentic
+retrieval baseline on corpus A, both on the same base model, both graded by the
+grader in this repository. The self-model answers better on every question type
+and pays about a thousand times more to build its memory.
```
apurvvkumaria: Blocker: This remains an unsupported cost ratio. Both published reports set `comparable_on_cost` to false, and `results/README.md` now correctly says that no multiple is established, but this README still claims “about a thousand times more.” Please remove or restate this sentence using only independently observed counts. Also include this top-level example README in the new no-cost-ratio regression; `_published_prose()` currently scans only `results/README.md` and run summaries, which is why this contradiction passes.

examples/tools/agent-memory-benchmark/bench/grader.py
apurvvkumaria: Rechecked at `f47a108c4894cd0d52ab633a91d82fb8c48008a9`: this remains reproducible. Both `This wrong estimate was corrected to 75%` and `This wrong value was replaced with 75%` return `correct=false` because `is wrong` is still matched across the `This wrong` token boundary. The latest commit does not change the grader or add these regression cases.

examples/tools/agent-memory-benchmark/README.md
apurvvkumaria: Rechecked at `f47a108c4894cd0d52ab633a91d82fb8c48008a9`: the root example README still says the self-model `pays about a thousand times more`, while both reports mark cost as non-comparable. `_published_prose()` still scans only `results/README.md` and the run summaries, so the new regression passes without checking this contradictory root README claim.

examples/tools/agent-memory-benchmark/results/README.md
```diff
@@ -0,0 +1,176 @@
...
+counts with a null price rather than inventing one. Read the tables in tokens,
+not dollars — the one priced phase is the cheapest thing either run did.
+
+Two rows go the other way, and they are in the table above rather than left
+out of it. The retrieval baseline refuses to answer more reliably: it abstains
+correctly on every question where the corpus does not support an answer, while
+the self-model answers from its record in cases where the record does not
+actually establish the fact. It is also slightly better at plain single-hop
+lookup, where there is nothing to synthesise and the extra step through a
+consolidated record can only lose detail. A design that reasons at ingest time
```
apurvvkumaria: Blocker: Adding the two loss rows is correct, but this paragraph turns one non-reproducible run into causal architectural conclusions. The published artifacts do not include the self-model adapter or its memory, so they cannot establish that the record caused the abstention misses, that consolidation `can only lose detail`, or that ingest-time reasoning generally buys and pays these tradeoffs. Please state only the observed differences for this run and label any explanation as a hypothesis. Also strengthen `test_the_results_page_names_every_type_the_self_model_loses` to require the losing type name or table row; checking only that both percentage strings appear somewhere in the README can pass without naming the type.

rebelle5868:
Both blockers are fixed at `a9b7320`, one of them was costing a real point on a published run, and checking the rest of the prose against the artifacts found four more claims that were wrong. One published number moves as a result.

**Token boundaries.** You were right that the comment claimed an anchoring the code did not do — the cues were compared with `in`, so any of them could fire inside a longer word. `This wrong estimate was corrected to 75%` reproduced exactly as you described.

It was not only a synthetic defect. Following it into the published answers found `is not` matching inside **`is noted`**: the self-model answer to `fact-020` reads *"Jin Oh (Jin) is noted as representing the ML-research side…"*, the whole clause was marked denied, and the question graded wrong for an answer that is plainly right. That cue shipped in the first version of this grader, so the point had been lost for as long as the file existed.

All three places that asked "is there a denial here" — the clause tagger, the positional check, and the set-denial test that governs a colon — now go through one boundary-aware table so they cannot drift apart again. Your two shapes and the `is noted` case are in the contract table, with a case asserting a real cue still denies.

**One published number moves.** Regrading takes the self-model run from 0.8925 to **0.8978**, its multi-source type from 93.2% to **94.5%**, and the headline difference from +6.5 pp to **+7.0 pp**. The agentic run is unchanged. `results/README.md` and the PR description carry the new numbers.

Regrading also exposed a gap in `tools/regrade.py`: it rewrote every score but left `provenance_note.effect` describing the previous grader, which is what the effect test kept catching by hand. Regrade now recomputes both halves from the two answer files.

**The root README's cost ratio.** Fixed, and the guard failed twice over exactly as you said: `_published_prose()` scanned only `results/README.md` and the run summaries, and its pattern required a numeral, which "a thousand times more" does not have. It scans every published page now — root README, results README, `docs/*.md`, run summaries — and matches on the phrase rather than a digit. Red-checked: it names `README.md:309` and quotes `'times more'`.

**Four more the same check found.** Having been wrong twice about prose the artifacts contradict, I read the rest of it rather than waiting for the next round.

1. It said no dollar figure was reported for either run. One was: `bench/pricing.py` prices the embedding model the agentic baseline uses at ingest, so that phase carried $0.0034. A test now requires any price a report carries to appear in the prose describing it.
2. It enumerated the publication rename as five singular identifiers when the map holds twenty-one text substitutions and four question ids. The counts are stated now, pinned against the map.
3. **It said the self-model answers better on every question type. It does not.** The retrieval baseline abstains correctly on all 13 questions the corpus cannot support where the self-model manages 10, and it is better at plain single-hop lookup, 86.7% against 83.3%. The results table did not contradict that sentence because it listed seven types and neither of these was among them — a selective result rather than a result. Both rows are in the table now, and a test requires every type the self-model loses to appear on the page.
4. It said generated run state stays out of version control. It did not: only `state/` and two side files were ignored, so a run directory was untracked rather than ignored and `git add -A` after a run would commit one. Both halves are asserted now — a fresh run is ignored, the two published runs are not.

The screenshot had drifted too, so it is checked rather than drawn: a test runs the self-test and requires every result line the SVG claims to appear in the output.

**A contamination trap, and what it cost.** Someone asked why corpus B has a README and corpus A does not. I wrote one, put it beside corpus A's documents, and the fingerprint test went red. The cause is worse than a fingerprint: adapters are handed the corpus directory and the shipped baselines walk it with `rglob("*.md")`, so that README became **document 426**. A third-party adapter cannot be assumed to skip it either. Anything inside a corpus directory is data. Corpus B never hit this because it already had a wrapper directory, with its prose and its questions outside `corpus_b/corpus/`.

Corpus A has the same shape now: `corpus/` → `corpus_a/corpus/`, `questions/` and `gold/` merged into `corpus_a/questions/` beside the answer key, and the page at `corpus_a/README.md`, one level above the documents. `gold/` as a name is gone. Adding a third corpus is copying a directory rather than inventing another set of paths.

**No score moves from the restructure.** `hash_tree` hashes paths relative to the corpus root and `hash_file` hashes content, so neither the rename nor the move changes a fingerprint, and the reports never stored a path. The comparability tests pass untouched, which is the evidence rather than my assurance. A test now asserts both corpus directories contain only corpus — `part_a`, `part_b`, the manifest, the counts, the canary, with `.md` only under `email/`, `slack/` or `chat/`. It fails on exactly the README that started this.

**The READMEs, and where they deviate from this repository.** They are reordered so a reader meets the demo, then the corpus, then the questions, then how to run it, with reference material after. I also added YAML frontmatter, badges, a table of contents and commented directory trees, to make them navigable by an agent as well as a person.

You should know that **no other example README here uses any of those** — frontmatter, badges, emoji headings and a table of contents are each zero out of the thirty-two. The required sections are intact: `## At A Glance` is complete but moved to the end as a reference card, with the rows that change what a reader does repeated where they act on them — notably the external-access disclosure, which now sits directly above the command that spends money, as the authoring rules ask. The screenshot section is titled "What A Run Looks Like" and now carries the terminal output as searchable text alongside the image.

If you would rather this matched the house style, every one of those is a cheap revert and I will make it. I would rather ask than have a formatting preference hold up the fixes above.

The suite is 205 offline tests. The PR description is updated.

rebelle5868:
Both parts fixed at `8601f66`. I missed this review when I replied a moment ago — it landed between your two earlier ones and my answer, and it did not cover it. Apologies.

**The paragraph was doing what I had spent the round objecting to elsewhere.** It took one non-reproducible pair of runs and stated causes: that answering from a record produced the abstention misses, that a consolidation step `can only lose detail`, and that ingest-time reasoning buys and pays these tradeoffs in general. These artifacts do not include the self-model's adapter or the memory it built, so nothing in them can show what any wrong answer was derived from. That the sentence flattered a reading I already had is exactly why it should not have been written.

It now states the counts and stops: 13 of 13 against 10 of 13 on `abstention`, 86.7% against 83.3% on `single_hop`, both visible per question in each run's `verdicts.jsonl`. The explanation that follows is labelled a hypothesis, says these two runs can neither confirm nor refute it, and says what testing it would take — runs where the memory is inspectable, on more than one corpus and more than one base model. The root README's summary is scoped the same way, and the same sweep found no other sentence generalising from these runs.

**The test could pass without naming the type**, as you said — two percentage strings anywhere on the page satisfied it. It now requires a table row that both names the type and carries both numbers. To make that findable, each type row in the table names the machine-readable type beside the human label, so `abstention` and `single_hop` appear as `| Refusing to answer when the corpus cannot — \`abstention\` (13) | 100.0% | 76.9% | -23.1 pp |`.

Red-checked in both directions: deleting the row fails while the percentages are still in the prose, and keeping the row but dropping the type name also fails. The first is the case you described.

205 offline tests.

examples/tools/agent-memory-benchmark/bench/grader.py
rebelle5868: Fixed at `8601f66`. The fallback is gone: `asserts()` now splits the answer into clauses and requires the value to appear in a clause that is not denied, with `contains()` running per clause so normalisation and denial scope see the same text. `None of these dates is correct: July 14, 2026` is denied through the normalised date, and `Sofia Ramos rather than Kofi approved it` asserts Sofia while denying Kofi. The generated attack and thirty-one hand-written shapes are in `DENIAL_SCOPE_CONTRACT`, both directions — an attack that must not score, and a valid correction that must.

examples/tools/agent-memory-benchmark/tests/test_published_results.py
rebelle5868: Fixed at `8601f66`. Both reports carry `schema_version`, `trial`, `run_parameters` describing their own run, and a structured `accounting` record. `adapter.revision` hashes the adapter tree and the shared lib for the agentic run, with a test asserting the recorded hash equals the adapter that ships; the self-model keeps a null hash and a note saying no adapter ships here. `test_the_report_follows_the_current_schema` enforces the current shape rather than accepting legacy output, and `test_accounting_claims_only_what_the_artifact_establishes` pins `forwarded_calls`/`uncounted_calls` null and `comparable_on_cost` false, which is what these legacy runs can support.

examples/tools/agent-memory-benchmark/tests/test_published_results.py
rebelle5868: Fixed at `8601f66`. `provenance_note.substitutions` now stores every substitution explicitly — twenty-one text entries and four question ids, in the order applied, with no "and three siblings". `test_the_published_map_reproduces_the_transformed_answers` applies that exact map to `answers.as-answered.jsonl` and compares the complete transformed rows against `answers.jsonl`, and `test_the_transform_changed_only_what_the_map_touches` asserts nothing outside the map moved. `results/README.md` states the map's size, pinned against the reports by `test_the_readme_states_the_size_of_the_substitution_map`.

examples/tools/agent-memory-benchmark/.gitignore
rebelle5868: Fixed at `8601f66`, and the prose was only half of it. The README now describes the real rule — generated run state stays out of version control, a result someone chose to publish is added explicitly — but `.gitignore` did not implement it: only `state/` and two side files were ignored, so a run directory was *untracked*, not ignored, and `git add -A` after a run would have committed one. I hit that while running the documented self-test. It now ignores `results/runs/*` and un-ignores the two published directories by name, and `test_a_generated_run_is_ignored_and_a_published_one_is_not` asserts both halves.

examples/tools/agent-memory-benchmark/results/runs/agentic-rag_corpus-a_nemotron/report.json
rebelle5868: Fixed at `8601f66`. The values are derived rather than restated: `test_the_published_effect_is_what_regrading_actually_gives` regrades both `answers.as-answered.jsonl` and `answers.jsonl` at the current head and asserts each against `provenance_note.effect`, so they cannot describe an older grader again.

Regrading also exposed why they drifted in the first place — `tools/regrade.py` rewrote every score but left `effect` untouched, which is why the test kept catching it by hand. Regrade now recomputes both halves from the two answer files. Current values are `0.8011 -> 0.8280` for agentic and `0.8710 -> 0.8978` for the self-model; they differ from the ones you computed because the grader moved twice more after your comment.

examples/tools/agent-memory-benchmark/bench/grader.py
rebelle5868: Fixed at `8601f66`, and it was not only a synthetic defect. Following your example into the published answers found `is not` matching inside **`is noted`**: the self-model's answer to `fact-020` reads *"Jin Oh (Jin) is noted as representing the ML-research side…"*, the clause was marked denied, and the question graded wrong for an answer that is plainly right. That cue shipped in the first version of this grader.

All three places that ask "is there a denial here" — the clause tagger, `_denied_at`, and the set-denial test that governs a colon — go through one boundary-aware table now, so they cannot drift apart. Your two shapes and the `is noted` case are in the contract table, with a case asserting a real cue still denies. The self-model run moves from 0.8925 to 0.8978 as a result.

examples/tools/agent-memory-benchmark/README.md
rebelle5868: Fixed at `8601f66`, and you were right that the guard failed twice over. `_published_prose()` scanned only `results/README.md` and the run summaries, and the pattern required a numeral, which "about a thousand times more" does not have. It scans every published page now — root README, results README, `docs/*.md`, run summaries — and anchors on the phrase rather than a digit. Red-checked against the old wording: it reported `README.md:309` and quoted `'times more'`. The sentence states where each run spends instead, with no multiple between the columns.

examples/tools/agent-memory-benchmark/results/README.md
rebelle5868: Fixed at `8601f66`. You are right, and it was the thing I had spent the round objecting to elsewhere: one non-reproducible pair of runs stated as causes. These artifacts do not include the self-model's adapter or the memory it built, so nothing in them can show what a wrong answer was derived from.

It states the counts and stops — 13 of 13 against 10 of 13 on `abstention`, 86.7% against 83.3% on `single_hop`, both visible per question in `verdicts.jsonl` — and the explanation after it is labelled a hypothesis these two runs can neither confirm nor refute, with what testing it would need. The root README's summary is scoped the same way.

On the test: it now requires a table row that names the type *and* carries both numbers, and each type row names the machine-readable type beside the human label so the row is findable. Red-checked both ways — deleting the row fails while the percentages remain in the prose, and keeping the row but dropping the type name also fails.

examples/tools/agent-memory-benchmark/results/runs/agentic-rag_corpus-a_nemotron/answers.as-answered.jsonl
```diff
@@ -0,0 +1,186 @@
...
+{"answer": "Alex said the provider's prefix cache works on hash equality — \"the prefix was still hashing the same on our side\" when the provider rolled their cache key format.", "source_ids": ["S:D200ROH001_dm@2026-05-21"], "id": "fact-003"}
+{"answer": "Section-level trimming was chosen (per-section basis using section headers in `prefix-stable`), not per-token.", "source_ids": ["E:2026-05-17T08-57-00__0e29129c", "E:2026-05-17T21-05-00__735ea58b"], "id": "fact-004"}
+{"answer": "May 29, 2026", "source_ids": ["E:2026-05-19T10-18-00__7a77c9c1"], "id": "fact-005"}
+{"answer": "O(|item_tokens|) — the select() method performs lookups proportional to the number of tokens in the item after the inverted index is built.", "source_ids": ["E:2026-05-18T14-22-00__87841ee4", "E:2026-05-19T09-30-00__2e5b9110"], "id": "fact-006"}
+{"answer": "MR !318 (merge request 318) at https://gitlab.example-co.com/platform-eng/atlas/-/merge_requests/318", "source_ids": ["S:D200ROH001_dm@2026-04-27"], "id": "fact-007"}
```
apurvvkumaria: **Blocker:** The published `answers.as-answered.jsonl` artifacts reintroduce the registrable domain that the corpus deliberately sanitized. This line contains `https://gitlab.example-co.com/...`; the self-model raw answers also contain that host and `morgan.lee@example-co.com`, and both reports publish `example-co.com` in the reversible substitution map. `docs/provenance.md` explicitly says this `.com` is registered and was replaced so published URLs use RFC 2606-reserved domains. Shipping the pre-publication files therefore defeats that boundary and contradicts the release/compliance claim that public content uses sanitized placeholders. Please regenerate the reference runs against the public corpus, or remove/redesign the raw artifacts and reversible map so no registered domain is published. Add a regression that scans every tracked result artifact, not only the corpus, for non-reserved domains.

examples/tools/agent-memory-benchmark/results/runs/self-model_corpus-a_nemotron/answers.as-answered.jsonl
```diff
@@ -0,0 +1,186 @@
...
+{"id": "fact-028", "answer": "Sam Iverson", "source_ids": ["self_model/raw/slack/C200ATLD001_channel/current.log"]}
+{"id": "fact-029", "answer": "Tomas Bergman", "source_ids": ["S:D200MOR001_dm@2026-04-26T13-06-24Z", "S:C200PFE001_channel@2026-04-26T13-47-40Z"]}
+{"id": "fact-030", "answer": "The corpus does not contain any statement where Alex Chen says he would put Wen-Qing on a specific workstream. The closest references are: (1) In email E:2026-05-21T09-24-00__88129152, Alex mentions an \"eval workstream\" but says they need \"one mid-level engineer owning eval\" — not Wen-Qing specifically — and notes Wen-Qing would be \"Q3 productive on ingest backfill\" and \"Q4 at earliest\" for end-to-end ownership. (2) Morgan's onboarding announcement (2026-04-27) states Wen-Qing will pair with Jordan on \"Atlas ingest paths\" for 6 weeks targeting independent sub-area ownership by end of Q3.", "source_ids": ["E:2026-05-21T09-24-00__88129152", "E:2026-05-23T08-41-00__a917fd3a", "/home/yijiel/gitrepo/mnemo-benchmark/results/runs/cos_a_nemotron/state/cos_home/workspace/self_model/people/wen_qing_tang.md"]}
```
apurvvkumaria: **Blocker:** This published `source_ids` entry exposes an absolute local workspace path under `/home/<username>/.../workspace/...`, including a user-specific directory. The same value is duplicated in `answers.jsonl` and `verdicts.jsonl`. That violates the public-repository rule against private workspace details and contradicts the PR compliance declaration. Replace it with a sanitized corpus-relative identifier in all three artifacts, then add a regression that scans tracked result files for absolute home/workspace paths so the disclosure cannot recur.

rebelle5868:
I ran a documentation review over every page this example ships, checking each falsifiable sentence against the artifact that would settle it rather than reading for plausibility. It found fifteen statements the artifacts contradict and about twenty that asserted more than they support. All are fixed at `96b9241`. Two of the fifteen are numbers on the results page, and one is something I told you in `a9b7320`.

**Correcting that first.** I wrote that a README placed beside corpus A's documents *"became document 426"*, and used that to justify the restructure. It would not have. `bench/runner.py:369` hands adapters `corpus_a/corpus/part_a` and then `part_b`, never the corpus root, so a file there is outside both `rglob` scopes — and the shipped adapters skip a file with no `doc_id:` anyway. I checked the history: that loop has been there since `b941efe`, so the claim was wrong when I made it, not invalidated by the move.

The mechanism is the one I did not look for. `bench.fingerprint.hash_tree` hashes **every file** under the corpus root — 428, not the 425 documents — so the README changed the corpus hash and every published result stopped describing what ships. Planting one moves it from `12772a52` to `6435d9db`. The restructure stands and the third-party-adapter caveat stands; the reason I gave for them did not. The page and `test_corpus_integrity.py`'s docstring, which repeated the same wrong mechanism, both say this now.

**Two published numbers were wrong, both from rounding a rounded figure.** The multi-source difference shipped as +6.9 pp; from the verdicts it is 5/73 = 6.8493, so **+6.8**. Citation coverage shipped as 97.9%; 182/186 displays as **97.8%**, and 97.9 came from the report's four-place `0.9785`. Nothing checked any cell in that table — including the first version of the test I wrote to check it, which read the same rounded field and duly agreed with the wrong number. It reads the counts now, and I re-derived every other cell: the rest reproduce.

**Every one of the root README's sixteen contents links pointed at nothing.** They used bare anchors while the headings carry a leading emoji, which GitHub renders as a leading hyphen; the body links in the same file already had it right. A test now resolves every in-page link on every page against that page's headings.

**Four things a submitter would have followed into a failure.** `name` is required and its absence is an uncaught `KeyError`, while `model`, which the interface marked required, is optional. The answer phase receives `{state}` and `{questions}` and *not* `{corpus}` — an adapter using `{corpus}` there dies after both ingest calls have already run and been paid for. Declaring `local` while routing through the proxy invalidates a run, where the page said such a run stays valid. And one blank or missing answer invalidates the whole submission, which the page did not mention anywhere; a blank answer is not credited as an abstention either. The last two are in Troubleshooting now.

**`docs/methodology.md` described a grader that does not exist.** It said hops are counted on the cost axis — nothing counts hops. It described `evidence_recall` as measuring what a system touched; it measures what a system reports in `source_ids`, against corpus `doc_id`s, so a store emitting its own ids scores zero until it maps them. It said `boolean` grading reads `accept`, which `_grade_boolean` never does. It gave the `string_any` rules as containment when all three run through the denial-scope check you had me build. And its freshness fallback describes a distinction the code cannot make, since a missing `stale_in_corpus` is indistinguishable from `false`.

**The publication substitution reached the answer text but not `source_ids`**, so 114 published self-model rows kept a pre-rename path while the answer beside them did not — and the rebuild test shared the same blind spot, which is why it passed. Both cover `source_ids` now; regrading confirms no score moves. Stale `corpus/` paths from the move also survived in `provenance.md` and `SUBMITTING.md`.

**The rest is scope, and it is the same defect you found on the loss rows.** The results page described the self-model's memory as a consolidated record it answers from — two sections above the sentence saying its memory does not ship. Those cells report call and token counts now and say the mechanism is not observable here. "One thing differs" is not what the reports show: the baseline is a round-capped configuration and calls a second model from another provider. The canary was called a tripwire when nothing watches for it. Corpus B is not "the same format as corpus A" — its emails carry 8 of corpus A's 14 fields, and its chat directory and `source` value differ, which silently costs a path-hardcoded adapter the whole chat half. And corpus A's page claimed corpus B was cleaner by construction, which the disclosed defects do not bear out in either direction.

217 offline tests. The PR description is updated, including the two corrected figures.

examples/tools/agent-memory-benchmark/results/runs/self-model_corpus-a_nemotron/answers.as-answered.jsonl
apurvvkumaria: Rechecked at `96b924113a334bd94c09080379d11eb8fbfd6f2e`: this blocker remains. The latest substitution changes the suffix from `self_model/...` to `wiki/...` in `answers.jsonl` and `verdicts.jsonl`, but all three published artifacts still contain the absolute `/home/yijiel/.../workspace/...` path; `answers.as-answered.jsonl` retains the original suffix as well. Please replace the absolute prefix with a sanitized corpus-relative identifier in all three artifacts and add a regression that scans every tracked result artifact for absolute home or workspace paths. This is the only substantive blocker remaining from my current review; the 217-test suite and required checks pass.

examples/tools/agent-memory-benchmark/results/runs/agentic-rag_corpus-a_nemotron/answers.as-answered.jsonl
rebelle5868: Fixed at `9b1fd1c` (originally at `686b48f`); replying here because I answered this in a top-level comment and never in the thread.

You were right about the shape of the failure: the sanitization covered the corpus, and `answers.as-answered.jsonl` is by definition the text from before it, so publishing that file put the registrable `.com` back — once in the agentic run, five times in the self-model run — while both reports listed it in a map written to be reversible.

I took the second of your two options. That one substitution is now pre-applied in both raw artifacts and is deliberately absent from the published map, so nothing here carries the domain and nothing here can reconstruct it. The map still reproduces `answers.jsonl` from the raw file row for row, because the substitution it no longer lists has already been made in the source. `results/README.md` states that a further substitution was applied and not listed, and why. The prose that explained the sanitization named the domain to make its point; it does not any more, since naming it in a public document is the same act as shipping it in one.

The regression scans every tracked file, not only the corpus, in three scopes so a failure says where. It reads hosts from URLs and addresses everywhere and bare hosts from prose only — `team.info` is a Slack API method quoted in a corpus rate-limit table, and `os.environ.get` is a call. Real infrastructure and inspected non-domains are separate lists: the first asserts we talk to it, the second asserts someone read it in place.

That guard is also what made the absolute-path leak you found next legible as the same class of defect rather than a one-off.

examples/tools/agent-memory-benchmark/results/runs/self-model_corpus-a_nemotron/answers.as-answered.jsonl
rebelle5868: Fixed at `9b1fd1c`, and you undersold it: the string carried a username, a directory layout and an internal project name at once — `/home/<user>/gitrepo/<project>/results/runs/<run>/state/<home>/workspace/...` — in all three artifacts of that run. Same class as the domain, and the reason is the same one: the sanitization covered the corpus and the answer *text*, and an artifact was published that predates both.

The prefix is gone from both answer files and regrading rebuilt the verdicts, so the citation reads like the 113 beside it: `self_model/people/...` before the publication rename, `wiki/people/...` after. The map still reproduces `answers.jsonl` from the raw file row for row. No score moves — a memory page name is not a corpus `doc_id`, so it scored zero either way, which the regrade confirms.

The regression scans every tracked file for absolute home paths and for a `~/` shorthand, in two scopes so a failure says where. Two exceptions, both explicit: the README's worked example of an env override is a `~/` path written for a reader rather than recorded from a machine, and the file that spells the patterns out is skipped because it matches itself. Red-checked by putting a `/home/...` citation back.

Looking for the same defect elsewhere turned up one more you had not flagged: `run_id` was still `cos_a_nemotron`, naming the system this publication calls self-model everywhere else. It is `self_model_a_nemotron` now.

I tried putting that rename in `provenance_note.substitutions` first and the map-size test went red, correctly — the two reports disagreed, 21 against 20. That map exists to reproduce `answers.jsonl` from the raw rows, and a `run_id` names no row, so adding it would have broken what the map means. It is recorded in `provenance_note.run_id_note` and disclosed on the results page instead.

219 offline tests.

apurvvkumaria (approved):
All substantive review findings are addressed at `9b1fd1cc0414eb2e06fa1119c8951d61ab1e6495`. The published artifacts now preserve the public-repository boundary, the regression coverage prevents recurrence, all 219 offline tests pass, and the repository checks are clean.

PR #145, opened 2026-08-27
feat(memory-driven-chief-of-staff): join a person's identities on the user's word

rebelle5868:
## Related Issue

Related: #122

## Description

A person is a set of identities, not one. `sender_key` from #140 gave a stable identity per source and nothing joined them, so a colleague who writes from Slack and from mail held two pages with half their history each — and neither got the readable name, because two identities contending for one slug both take a digest. That split becomes visible the moment a second connector is configured, which #142 made possible.

**Only the user joins identities, and the job never waits for the answer.** A matching display name is not evidence — guessing from one is the mistake `sender_key` exists to prevent — and a matching handle is a reason to ask rather than to write. The memory job reports `identity_candidates`, writes a page per identity as before, and exits. The user answers whenever they get to it through `link_identity.py`, which is the only thing that records an answer, for the same reason `correct.py` is the only thing that writes `actor='user'`: a system that can manufacture its own evidence will eventually believe something nobody told it. A nightly job whose completion depends on when somebody read a message is a job that does not complete.

**Answers are stored pairwise and resolved with a disjoint set, so the design does not count to two.** Confirming A~B and B~C joins three identities without a third question, and a fourth connector is one more pair rather than a new shape. Rejections are recorded as durably as confirmations, because a candidate re-proposed on every run is how an idle job wakes the agent forever. When answers stop agreeing — a rejected pair joined anyway by later confirmations — that is reported as `identity_conflicts` and nothing is changed, because neither side is knowably the wrong one.

**An identity is a `(source, key)` pair, never a bare key.** Two systems can mint the same string and merging their owners would be silent and unrecoverable. The text form splits on the first colon only, because a key may contain colons of its own — a Teams conversation id is `29:1a2b…`.

**Schema v5 makes a new connector an INSERT rather than a migration.** `CHECK (source IN ('email','slack'))` meant a third connector could not write a single row until somebody shipped a schema change, which is a poor thing to discover while writing the connector. `sources` is a table and `items.source` a foreign key; the constraint still refuses a typo, with the caveat — stated in the schema — that SQLite enforces foreign keys per connection, so every connection this code opens for writing now sets the pragma. Two did not.

**`sender_handle` records a third thing, distinct from who somebody is and from what they are displayed as.** A Slack handle, a login, a mail local part: unique within its source, changeable by its owner, and the value a user recognises when asked whether two identities are the same colleague. Slack's `users.info` already returned it beside the display name and it was being discarded, so this costs no request. For mail it is derived from the address rather than stated, which the comment says, because it changes what the value is worth as evidence.

**People pages carry `identities:` — a list.** Privileging one entry as `source_key` made "which is the primary" a question with no answer as soon as a third arrived. Pages written before the list are read as a list of one and are not rewritten; the page belongs to the user. `memory_check` gains a malformed-entry finding beside missing and duplicate, because an entry with no source matches nothing and orphans the page silently rather than breaking it loudly.

One defect is worth naming because verification found it and the tests did not. The v5 step drops and recreates `items`, and `obligations.source_id` references it `ON DELETE CASCADE` — with foreign keys enabled SQLite performs an implicit `DELETE FROM` before dropping a table, which fired that cascade and the cascade from `obligations` to `events` after it. A v4 store holding one message, one obligation and one event came out of the migration with the message alone. The row-count guard missed it because it counted `items`, the one table that survived. `PRAGMA foreign_keys = OFF` does not help, because SQLite ignores it inside a transaction and migrations run inside one; `defer_foreign_keys` defers constraint checks rather than cascade actions. Both were measured. The dependants are now carried across the rebuild, discovered as the transitive closure of `PRAGMA foreign_key_list` rather than named — a first attempt saved only the direct dependants, restored the obligations and lost every event.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — 572 files
- [x] `git diff --check`
- [x] Recipe suite: 603 tests across fourteen files, `OK`
- [x] Both connectors end to end through the real normalizers on a fresh store: eight messages in as three mail and five Slack; before an answer, two `Dana Okoro` pages with one identity each and one `shared handle` candidate; after `link_identity.py same`, one `dana_okoro` page holding all six of her messages, no candidates and no conflicts; a Slack-only colleague unaffected throughout
- [x] A third connector, registered by `INSERT INTO sources`, wrote a row with a colon-bearing key and joined the existing candidate group as one question rather than re-asking the pair already answered
- [x] `schema-v3.sql` and `schema-v4.sql` upgraded to v5 with `items`, `obligations`, `events` and `cursors` preserved row for row and `PRAGMA foreign_key_check` clean
- [x] The offline fixture walkthrough runs clean end to end, including `memory_check` reporting zero findings and reporting one when a required field is removed on purpose
- [x] Each behaviour checked by reverting it and confirming the assertions turn red — dropping the source from an identity, resolving groups from direct pairs only, reading a rejection as an answer, and both stages of the migration defect
- [ ] Not verified against a live Hermes install; the connectors' own paths are unchanged by this PR and the store changes are exercised against the frozen schema artifacts instead

- Result: `docs-updated`
- Evidence or justification: `README.md` — the identity section replaces the per-source split note with the proposal-and-answer flow and the command; `profile/schema.md` — `identities` replaces `source_key` in the people-page contract, with the reason it is a list and the rule that only the user adds an entry; `profile/skills/memory-writing/SKILL.md` — how to copy the list, when to raise a candidate in conversation, and an explicit instruction not to wait for the answer.
- Reviewer: @rebelle5868

## Release And Compliance

### DIFF
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/link_identity.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/link_identity.py
@@ -80,16 +80,27 @@ def _check_sources(conn, people: list[identity.Identity]) -> None:
 
 
 def decide(texts: list[str], status: str) -> dict[str, object]:
-    """Record one answer over every pair among the named identities.
+    """Record the user's answer, pairwise.
 
-    Stored pairwise even when the user answered about a group, so a later
-    answer about one member does not have to unpick a group decision — and so
-    `different` means what it says. Confirming a group of three is three
-    pairs; rejecting one is also three, because "these are not one person"
-    does not say which of them is the odd one out, and recording only some of
-    the pairs would leave the rest to be asked again.
+    `same` takes any number, because "these are one person" is a claim about
+    every pair in the group and each of them follows from it. Confirming
+    three identities is three pairs, and a later answer about one member can
+    revise that pair without unpicking a group decision.
+
+    `different` takes exactly two, because the negative does not distribute.
+    "These three are not one person" rules out the group and says nothing
+    about which member is the odd one out — A and B may well be the same
+    colleague with C a stranger. Recording all three pairs as rejected would
+    bury the A~B link and stop it ever being proposed again, on the strength
+    of an answer the user did not give.
     """
     people = _identities(texts)
+    if status == "rejected" and len(people) != 2:
+        raise ValueError(
+            "`different` takes exactly two identities. Saying a group is not "
+            "one person does not say which member is the odd one out, and "
+            "recording every pair as rejected would bury a link the user "
+            "never denied. Answer the pairs that are actually different.")
     unknown = []
     with write_txn() as conn:
         _check_sources(conn, people)
@@ -143,7 +154,7 @@ def main(argv: list[str] | None = None) -> int:
     same = sub.add_parser("same", help="one person, several accounts")
     same.add_argument("identities", nargs="+", metavar="source:key")
 
-    diff = sub.add_parser("different", help="not the same person")
+    diff = sub.add_parser("different", help="these two are not one person")
     diff.add_argument("identities", nargs="+", metavar="source:key")
 
     shown = sub.add_parser("list", help="what has been answered so far")
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/link_identity.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/link_identity.py
@@ -1,0 +1,170 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""The user's answer to "are these the same person?".
+
+Like `correct.py`, and for the same reason: the model must not be able to
+manufacture this evidence. A display name matching, a handle matching — those
+are why a question is asked, never why it is answered. Only the user answers,
+and this is the only thing that writes an answer.
+
+    python3 link_identity.py same slack:U01DANA email:dana@example.com
+    python3 link_identity.py different slack:U01DANA email:d.okoro@example.com
+    python3 link_identity.py list [--all]
+
+Any number of identities may be named at once, because that is the shape of
+the question the memory job asks — "these three all present as Dana Okoro" —
+and answering it as three separate pairs is arithmetic the user should not
+have to do.
+
+**This is deliberately not part of any scheduled job.** The memory job writes
+its pages, records what it noticed, and exits; nothing waits for an answer,
+and an unanswered question costs nothing. Blocking a nightly job on a human
+would make the job's completion depend on when somebody happened to read a
+message, and the pages it can write without the answer are still worth
+writing.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from itertools import combinations
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+import identity  # noqa: E402
+from _db import ensure_store, write_txn  # noqa: E402
+
+
+def _identities(texts: list[str]) -> list[identity.Identity]:
+    seen = []
+    for text in texts:
+        who = identity.parse(text)
+        if who in seen:
+            raise ValueError(f"{who} was named twice")
+        seen.append(who)
+    if len(seen) < 2:
+        raise ValueError("name at least two identities")
+    return seen
+
+
+def _known(conn, who: identity.Identity) -> bool:
+    return conn.execute(
+        "SELECT 1 FROM items WHERE source = ? AND sender_key = ? LIMIT 1",
+        (who.source, who.key)).fetchone() is not None
+
+
+def _check_sources(conn, people: list[identity.Identity]) -> None:
+    """Refuse a source this install has never heard of.
+
+    The foreign key already refuses it, and says `FOREIGN KEY constraint
+    failed` — which tells a user who mistyped `slak:U01DANA` nothing about
+    what went wrong or what to type instead. The constraint stays; this is
+    the message.
+
+    Refusing rather than adding the source: a source appears when a connector
+    ships, and inventing one here would let a typo create a namespace that
+    silently matches nothing forever.
+    """
+    known = {row[0] for row in conn.execute("SELECT name FROM sources")}
+    missing = sorted({w.source for w in people} - known)
+    if missing:
+        raise ValueError(
+            f"no connector called {', '.join(missing)}; this install knows "
+            f"{', '.join(sorted(known))}. An identity is `<source>:<key>`, "
+            "and the source is the connector that collected it.")
+
+
+def decide(texts: list[str], status: str) -> dict[str, object]:
+    """Record one answer over every pair among the named identities.
+
+    Stored pairwise even when the user answered about a group, so a later
+    answer about one member does not have to unpick a group decision — and so
+    `different` means what it says. Confirming a group of three is three
+    pairs; rejecting one is also three, because "these are not one person"
+    does not say which of them is the odd one out, and recording only some of
+    the pairs would leave the rest to be asked again.
+    """
+    people = _identities(texts)
+    unknown = []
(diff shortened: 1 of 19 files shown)


examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -196,50 +197,63 @@ def slug(name: str) -> str:
...
+    claimed: dict[str, str] = {}
+    for mark, page in have.items():
+        for text in page.get("identities") or []:
+            claimed.setdefault(text, mark)
+
+    names: dict[object, str] = {}
+    for person in display:
+        for who in members.get(person, [person]):
+            if str(who) in claimed:
+                names[person] = claimed[str(who)]
```
apurvvkumaria: Blocking: this loop selects the first existing page that claims any identity in the newly joined group and drops the other matched pages from the handoff. The normal pre-confirmation state is one page per identity. After the user confirms two identities, both pages still exist, but only one slug is returned; the other page is not merged, removed, or reported. Its accumulated history and index entry are stranded, contrary to the stated outcome of one page holding the complete history. Please return all matched pages or explicit merge inputs, and add an end-to-end test that starts with two populated pages, confirms the link, and verifies that one retained page contains both histories while the other is handled deliberately.

apurvvkumaria (commented):
Two blocking identity-integrity findings are attached inline. Repository checks pass and the pull request is conflict-free, but these cases can leave durable person records incorrect or incomplete.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/link_identity.py
```diff
@@ -0,0 +1,170 @@
...
+    does not say which of them is the odd one out, and recording only some of
+    the pairs would leave the rest to be asked again.
+    """
+    people = _identities(texts)
+    unknown = []
+    with write_txn() as conn:
+        _check_sources(conn, people)
+        for who in people:
+            if not _known(conn, who):
+                unknown.append(str(who))
```
apurvvkumaria, line 109: Blocking: `different` over three or more identities records every pair as rejected, but a negative answer to “are these identities all one person?” establishes only that the whole group is not the same person. For example, A and B may be one person while C is different. Persisting all pairwise rejections incorrectly rejects A/B and suppresses that valid link from future candidates. Please restrict `different` to exactly two identities, or collect explicit pairwise decisions for a group, and add a regression test for this case.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Reproduced, and you are right that it contradicts the description I wrote. `page_names` took the first page claiming any of a person's identities and stopped, so after a confirmation the other page kept its history and its index entry and belonged to nobody — while the handoff claimed one page held everything.

It lives in the state this feature exists to resolve rather than the one my own end-to-end check started from — an empty memory, where each identity's page has not been written yet and neither of these can happen. That is the same gap that let three earlier defects through here, and it is worth naming again.

Every page a person owns is now returned, the extras as `merge_into_slug`. The kept page is chosen by sorted order so it does not move between runs: a keep-choice that changed would shuttle content back and forth and lose some each way. Nothing counts to two — three pages for one person are all reported.

The merge itself stays the agent's, and the skill now says what that means: fold each named page into the kept one and remove it, index entry included, and merge by hand rather than by concatenation, because Recent Interactions is newest-first with a ceiling and two lists spliced end to end are neither.

Regressions, each checked by reverting the fix: two already-populated pages plus a confirmation, the kept page stable across runs, the kept page carrying both halves of the history, three pages for one person, and an empty `merge_into_slug` when there is nothing to merge. Reverting turns two of them red.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/link_identity.py
rebelle5868: Correct — the negative does not distribute, and recording it as though it did buries a link the user never denied.

`different` now takes exactly two, and the error says why rather than only refusing: saying a group is not one person does not say which member is the odd one out, so answering the pairs that are actually different is the only shape that means anything. `same` still takes any number, because that claim does distribute over its pairs and each of them follows from it.

Two regressions: a group answered as `different` is refused with nothing recorded, and the pair the user did deny is still recorded so refusing the group does not refuse the answer they can give. There is also a third pinning that `same` over three still records its three pairs, so the restriction did not quietly land on both verbs.

Worth saying: I fixed this before writing its regression, and the revert probe came back green — which is how I found out the coverage was missing. It turns one red now.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
```diff
@@ -419,39 +505,58 @@ def evidence(conn, since: str) -> dict[str, object]:
...
         candidates.append({
             "sender": sender,
-            "source_key": key,
+            # Pages that belong to this person and are not the one above.
+            # Present only after a link is confirmed over identities that had
+            # each already been written up. Their content has to be folded
+            # into `slug` and the page then removed, index entry included —
+            # nothing else will do it, and left alone they are history
+            # attributed to nobody.
+            "merge_into_slug": extra.get(key) or [],
```
apurvvkumaria, line 522: Blocking: a pending page merge must bypass the quiet-page freshness filter above. If both pages already record the latest interaction when the user confirms their identities, line 504 continues before this field is emitted. I reproduced `people: []` and `identity_candidates: []` with two current populated pages still on disk after the link. Because no new message is required to arrive, those pages can remain split indefinitely and the top-level gate can sleep. Please treat `extra.get(key)` as work regardless of message freshness, and add a regression with both pages current that verifies `merge_into_slug` is emitted and the memory-writing job wakes even when attention pages and all other inputs are current.

apurvvkumaria (commented):
One remaining blocker is attached inline. The previous two findings are fixed, but a pending page merge can still be suppressed permanently by the quiet-page filter.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/select_memory.py
rebelle5868: Reproduced, and it is the interaction between the merge report and the rule it sits behind rather than either one alone. The freshness rule asks whether a message has arrived that the page does not already account for — right for a quiet correspondent, and exactly wrong here: both pages being current is the case where a split sits unnoticed, and no new message has to arrive for it to still be true. So the person was skipped, the field was never emitted, and the gate slept because it reads the same list.

A pending merge is now work regardless of freshness. The exemption is that and nothing more — a quiet person with nothing to merge still costs no model call, which has its own regression so the exemption does not quietly become a way around the rule. Three tests, each with every other input current: attention pages fresh, no corrections, and no message since either page was written, so the merge alone has to carry the wake. Reverting turns two of them red.

Two things came out of asking what the fix had made possible, rather than waiting to be told.

`duplicate-identity` said one of the two pages was about somebody else. After a merge that copies the content across and does not delete what it emptied, the same clash means the opposite, and the repair skill was reading the first meaning and leaving the leftover alone forever. I tried to separate them by identity set — a page whose identities are all on another is a leftover — and an existing test caught what that costs: a page that wrongly claims somebody else's address makes the victim's page look like the subset, so the advice becomes "delete this page" about the correct one. The separating evidence is the confirmed link, which lives in the store and not on disk. The finding now reports both readings and refuses to choose, and points at `merge_into_slug`, which is where the answer actually is. A test pins that the detail carries no deletion advice.

The README also now records that a digest page name is permanent. Two people presenting one display name each get one; confirming they are one person removes the ambiguity but not the name, because renaming means rewriting the index entry and every link pointing at it — a real risk taken for an appearance, and a page name that moves is how history gets lost.

603 tests.

apurvvkumaria (approved):
All substantive review findings are addressed at 101fdc6ab6dd604684d6b6c10865d0f1bcaa0698. The identity-merge handoff preserves all pages, grouped negative answers no longer invent pairwise rejections, pending merges bypass the quiet-page gate, all 603 tests pass, and required checks are clean.

PR #153, opened 2026-08-28
fix(tools): keep published citations inside the corpus, and restructure the README

rebelle5868:
## Summary

Three changes to this example, in one PR because they touch the same published artifacts.

**Citations now name only what the corpus contains.** Two `source_ids` entries in the self-model run referenced repositories that are not part of the corpus. They are replaced with one that is, so every published citation points at published material. The replacement is pre-applied in the raw artifact and is not listed in the substitution map, following the rule the map already states: a map that lists a pre-image republishes it.

**A refusal is no longer scored as an assertion.** Saying the corpus is silent denies everything the sentence goes on to name, and the grader did not model that. The self-model's answer to `abstain-compass-cutover-done` reads *"there is no information in the corpus about whether it completed successfully"* — a correct refusal — and was marked wrong for asserting the phrase it was declining to assert. Separately, an answer that places an event in the future has not claimed it happened; the cue list carried the past forms (`only planned`, `has not happened`) but not the present ones a system actually writes.

**The README puts the reader first.** Demo, corpus, questions, getting started, results, troubleshooting — then a single `For Contributors` section holding the structure, the adapter contract, the isolation notes and the verification steps, so a reader who only wants to run the benchmark can stop before it. The transcript that repeated the screenshot line for line is replaced by one searchable line of the same result.

**The `Metadata` YAML block is gone from all four READMEs.** `CONTRIBUTING.md` asks that catalog data live in the Markdown table rather than in YAML, and most of what those blocks held was a second, untested copy of a fact the prose and the artifacts already carry — which is how several of the claims this branch fixes came to be wrong in the first place. The entry point, the offline check, the corpus locations and the adapter contract moved into Getting Started, where a reader needs them; nothing else in them was unique.

## Published numbers move

| | Before | After |
| --- | ---: | ---: |
| Self-model, overall | 89.8% | **90.9%** |
| Self-model, `abstention` | 76.9% (10/13) | **92.3% (12/13)** |
| Headline difference | +7.0 pp | **+8.1 pp** |

`fingerprint.gold` is unchanged — the answer key was not touched. Only `fingerprint.scorer` moves, and both runs were regraded against it. The citation change moves no score, since a repository id is not a corpus `doc_id`.

## What keeps each from recurring

- An `org/repo` used as a repository reference — followed by `!123`, `#123` or `/-/` — must be one the corpus contains. A bare `word/word` is ordinary prose and occurs in these answers, so the check is anchored on the reference form rather than on the slash.
- That check reads **whole rows**, not the `answer` field, so a citation is covered as well as the text beside it.
- Seventeen denial cues for the refusal shape, with four contract cases in both directions: widening a denial must not swallow a value asserted after it, so *"there is no information about the rollout; the figure is 50%"* still asserts `50%`.

Each check is verified by reinstating what it rejects and confirming it fails.

## Notes on two upstream interactions

- `scripts/build_catalog.py` requires the README title to be followed by the catalog table, so that table stays at the top; it is a contract rather than a layout choice.
- The same change moved the README image to a raster, because the catalog accepts only raster formats. A raster cannot be read back and compared against the run it depicts, so the drift check still reads the SVG. A test now requires the SVG to ship whenever the README shows an exported image, and `docs/assets/README.md` records which file is the source and which is the export.
- Repairs a check that was already failing on `main`: the example catalog reworded this example's row upstream, and the check was pinned to that sentence rather than to the number it was verifying.
- This branch merges `main` rather than rebasing, per `CONTRIBUTING.md`. The upstream README rewrite is taken as the base for all four files; what this branch changes is reapplied on top, including the published figures, since that rewrite branched before the refusal fix and carried the superseded ones.

## Verification

**Evidence level:** local/static.

- [x] `python3 -m pytest tests/` on a clean tree at `790b92c` — **234 passed, 0 skipped**, offline. `tesseract` is required by the two image checks and is listed in the prerequisites, the install block, troubleshooting, the contributor dependency note and the Requirements row; without it the suite reports 232 passed, 2 failed.
- [x] `python3 scripts/check_license_headers.py --check` — clean
- [x] No tracked file, commit message or description in this branch names a domain, path or repository outside the corpus

## Related

Follow-up to #143.

---


examples/tools/agent-memory-benchmark/README.md
```diff
@@ -17,48 +17,30 @@ across two synthetic corpora.
...
 - [Cleanup](#cleanup)
+- [For Contributors](#for-contributors)
+  - [Project Structure](#project-structure)
+  - [Configuration](#configuration)
+  - [Adding Your System](#adding-your-system)
+  - [How The Runner Treats Your Adapter](#how-the-runner-treats-your-adapter)
+  - [Verification](#verification)
 - [Provenance And License](#provenance-and-license)
-- [Metadata](#metadata)
+- [Catalog Metadata](#catalog-metadata)
```
apurvvkumaria: The catalog check currently fails because this TOC entry links to `#catalog-metadata`, while the generated catalog removes that metadata section from the rendered body. Remove this TOC entry (or otherwise make the fragment resolvable), then rerun `scripts/build_catalog.py --check`.

examples/tools/agent-memory-benchmark/bench/grader.py
```diff
@@ -140,6 +149,15 @@ def _is_abstention(answer: str) -> bool:
     "is not", "are not", "was not", "were not", "isn't", "aren't", "wasn't",
     "weren't", "no longer", "rather than", "instead of", "not ", "never ",
     "cannot", "can't", "don't", "doesn't", "didn't", "no evidence", "not the",
+    # Saying the corpus is silent is a denial of everything the sentence goes on
+    # to name. A correct abstention that read "there is no information about
+    # whether it completed successfully" was scored as asserting the phrase it
+    # was declining to assert. A bare "no" is too broad to be a cue; these are
+    # the shapes an abstention actually takes.
+    "no information", "no mention", "no record", "no indication", "no detail",
```
apurvvkumaria: These bare cues are too broad. For example, `There is no information gap and the figure is 50%.` and `There is no data issue and the figure is 50%.` are both graded as though `50%` were denied. Please narrow the silence cues to actual refusal constructions and add positive-assertion regression cases.

examples/tools/agent-memory-benchmark/tests/test_no_registrable_domains.py
```diff
@@ -155,3 +155,81 @@ def test_no_tracked_file_publishes_a_path_from_someone_s_machine(kind):
...
+# in these answers.
+_ORG_REPO = re.compile(
+    r"(?:[a-z0-9.-]+\.[a-z]{2,}/)?"
+    r"([a-z][a-z0-9_-]{2,}/[a-z][a-z0-9_.-]{2,})"
+    r"(?=[!#]\d|/-/)")
+
+# Repository paths this example legitimately names.
+KNOWN_REPOSITORIES = {
+    "platform-eng/atlas", "platform-eng/onboarding", "platform-eng/perf",
+    "platform-eng/compass", "platform-eng/quillon",
```
apurvvkumaria: This allowlist weakens the corpus-containment guarantee: two entries in the set do not occur as repository references in corpus A, so future citations to them would still pass. Please derive the allowed set from corpus references, or remove entries that are not actually present and add a negative regression.

examples/tools/agent-memory-benchmark/tests/test_no_registrable_domains.py
```diff
@@ -155,3 +155,81 @@ def test_no_tracked_file_publishes_a_path_from_someone_s_machine(kind):
...
+    assert not offenders, (
+        f"published results name repositories that are not part of the corpus: "
+        f"{offenders}. A model naming its own source repository in an answer is "
+        f"the corpus rename's blind spot, because the corpus never held the string.")
+
+
+def test_a_published_answer_names_only_things_the_corpus_contains():
+    """A model can name an internal thing the corpus never held.
+
+    `hwinf-afe/cos!342` and four `atlas/solswarm!NNN` ids reached published
```
apurvvkumaria: This regression-test documentation republishes the source identifier that the PR is intended to remove. Replace it with a clearly fictional fixture value; a public safeguard should not preserve a removed identifier in comments or test text.

examples/tools/agent-memory-benchmark/tests/test_selftest_end_to_end.py
```diff
@@ -138,3 +138,25 @@ def test_the_screenshot_shows_what_the_self_test_actually_prints(tmp_path):
...
+    means anything while the SVG still ships and still describes the same
+    output, which the check above enforces -- this one enforces that the pair
+    stays a pair.
+    """
+    readme = (REPO / "README.md").read_text(encoding="utf-8")
+    shown = re.findall(r"!\[[^\]]*\]\((docs/assets/[^)]+)\)", readme)
+    assert shown, "the README no longer shows the offline self-test image"
+    for reference in shown:
+        assert (REPO / reference).exists(), f"the README shows {reference}, missing"
+    if any(not r.endswith(".svg") for r in shown):
```
apurvvkumaria: This proves only that the raster and SVG both exist. The README-displayed PNG still shows 135 tests, while the updated SVG and documentation say 226. Please re-export the PNG from the updated SVG and make the guard verify meaningful source/export synchronization rather than existence alone.

apurvvkumaria (commented):
Thanks for resolving the merge conflict. At this head, the benchmark suite passes locally (226 tests), but the PR is still blocked: the DCO workflow requires a valid Signed-off-by line in the PR description, and catalog validation fails on the unresolved catalog-metadata fragment. The inline comments identify the remaining correctness and publication issues.

examples/tools/agent-memory-benchmark/README.md
rebelle5868: Fixed at `3649d7c`; `Validate catalog` passes. The entry is gone. The upstream version of this file did not link that section either — I regenerated the contents while regrouping and added a link the build cannot resolve, since the catalog strips that section from the rendered body.

examples/tools/agent-memory-benchmark/bench/grader.py
rebelle5868: Fixed at `3649d7c`. Both of your examples reproduced, and a third of the same shape: `No record keeping was needed; the figure is 50%.`

The cues were bare phrases, so any sentence containing one read as a refusal. They are patterns now, and the absence word has to be followed by something that says *what* is missing — a preposition introducing the subject, a complement clause, or a verb taken by "nothing in the corpus". `no information gap` and `no data issue` no longer match; `no information in the corpus about whether …` and `nothing in the corpus records …` still do, including when a qualifier sits between the absence word and its subject.

Five cases join the contract in both directions: three sentences that merely contain an absence word and must still assert, two refusals that must still deny. No published score moves — the shapes do not occur in the stored answers, so this closes a hole that could be constructed rather than one that was being hit.

examples/tools/agent-memory-benchmark/tests/test_no_registrable_domains.py
rebelle5868: Fixed at `3649d7c`. You are right, and it was worse than two entries: `platform-eng/compass` and `platform-eng/quillon` appear zero times in corpus A, and two others appear once each.

The set is derived from the corpus now, by reading repository references out of `corpus_a/corpus/` at test time. A hand-written list drifts from the corpus it is meant to describe, and an entry that is not in the corpus silently permits exactly the citation the check exists to reject.

examples/tools/agent-memory-benchmark/tests/test_no_registrable_domains.py
rebelle5868: Fixed at `3649d7c`. You are right, and this is the fourth place in this work where removing a name and then writing it into the explanation happened — the substitution map, a provenance sentence, a commit message, and now a guard's own comments.

The comments describe the shape without naming the instances, and the docstring says why: a safeguard that removes a string should not preserve it in its own text. The rule is easy to state and I have been slow to apply it consistently, so it is now written where the next person will read it.

examples/tools/agent-memory-benchmark/tests/test_selftest_end_to_end.py
rebelle5868: Fixed at `3649d7c`. You are right that existence proved nothing — the raster was showing 135 against the SVG's count, and the test passed.

The PNG is re-exported from the current SVG. The guard reads the raster with OCR and requires the test count the SVG states; where `tesseract` is not installed it falls back to requiring the raster to be no older than its source, which is weaker but still catches an SVG edited without a re-export. Red-checked by restoring the previous PNG: it reports that the image does not show the count the SVG states.

`docs/assets/README.md` records which file is the source and which is the export, since the catalog's format rule means the checked file and the displayed file can never be the same one.

apurvvkumaria: Thanks—the PNG itself is updated, and the OCR path performs a meaningful comparison. One blocker remains in the no-`tesseract` fallback: Git does not preserve source file modification times. In a fresh checkout, the SVG and PNG receive the same checkout timestamp, so `raster.stat().st_mtime >= svg.stat().st_mtime - 1` can pass even when a stale PNG is committed with a newer SVG. Please replace the timestamp fallback with deterministic content-based validation, or make the OCR dependency mandatory for this test. The other review findings are addressed.

rebelle5868: Fixed at `70ca5e3`. You are right, and I had flagged that exact property of git to myself when I first wrote a timestamp check, then reintroduced it as a fallback anyway. It is the second unsound version of this guard.

The export tool now writes the SVG's SHA-256 into a PNG `tEXt` chunk, and the test compares that record against the SVG it ships beside. The record travels with the raster, so the comparison is deterministic in any checkout and needs neither OCR nor timestamps.

Red-checked twice, the first time with `touch` setting both files to the same mtime so the previous approach would have passed: editing the SVG without re-exporting reports the raster was exported from a different file and prints both digests, and stripping the chunk reports that the raster records no source digest.

The OCR comparison stays as a second, independent check where `tesseract` is present, and skips where it is not. The two prove different things — a digest proves the raster came from this SVG, and only reading the picture proves the export rendered what the SVG says — so neither replaces the other.

`tools/export_selftest_image.py` does the render and the stamp in one step, and `docs/assets/README.md` documents it as the only supported way to update the image.

apurvvkumaria (commented):
Two merge blockers remain at head `70ca5e3360850dd7a533242432c1f982ece182c8`:

1. **The source/export guard can still accept stale image content.** `tools/export_selftest_image.py --stamp` attaches the current SVG digest to an existing PNG without rendering it. I reproduced this by changing the SVG result from `232 passed` to `999 passed`, removing the old `source-sha256` chunk, and running `--stamp`. The digest test then passed while the PNG IDAT SHA-256 remained unchanged (`97233e794e736537580418f00b5977c2701fa1f79f24306093cb00393e6e8215`). Also, `stamp()` appends instead of replacing an existing source stamp, so `--stamp` does not correctly update an already stamped PNG. Please remove the stamp-only path and stamp only bytes produced by a successful render, replace any existing keyword chunk, and add a regression proving stale pixel data cannot be accepted. If content cannot be checked deterministically, make the rendering or OCR dependency mandatory for this test.

2. **The verification result is stale.** The PR description reports `226 passed`; the current merge result against `main` runs `231 passed, 1 skipped`. Please rerun after the final fix and record the exact result, including the conditional OCR skip.

The branch is otherwise conflict-free, and the required GitHub checks and catalog checks are green.

rebelle5868:
Both fixed at `5fa6203`.

**The stamp-in-place path is gone.** You are right, and it was a hole I opened for convenience: a digest could be attached to a raster nobody re-rendered. A digest is now attached only to bytes returned by a successful render, and stamping replaces any existing record rather than appending a second one — the append behaviour you noticed would have left two chunks on an already-stamped file.

**The content check is required rather than conditional.** That hole was survivable only because reading the image caught it, and that check skipped wherever `tesseract` was absent — so the guard that mattered was the one a machine might not run. It fails now with an install hint instead of skipping. A digest records what an exporter claimed; only reading the picture proves the pixels were re-rendered. I had treated them as alternatives, and they are not.

A regression performs your forgery through the internal helper — stale bytes, current digest — and requires the content check to reject it. It also asserts no stamp-in-place entry point has returned. I re-ran your exact reproduction against the fix: `--stamp` no longer exists, and forcing the forgery through the helper turns the suite red.

**Verification, rerun on a clean tree at `5fa6203`:** `233 passed`, **no skips** — the conditional skip disappeared with the conditional check. The PR description now records that, and notes the `tesseract` requirement.

apurvvkumaria:
The original two blockers are addressed at head `5fa62039f932088e2b104b88730a85f848f0604f`. One merge blocker remains:

**The newly mandatory `tesseract` dependency is missing from the public setup instructions.** The README still says that the offline check needs only Python 3.9+ and `pytest`, and its printed setup installs only `pytest`. Following those instructions on a clean machine produces `231 passed, 2 failed` because the two image-content tests require `tesseract`.

Please update the main prerequisites, Getting Started installation steps, troubleshooting guidance, contributor dependency description, and Catalog Metadata requirements to include `tesseract` (with the macOS/Linux install commands), then rerun and record the complete result.

The required GitHub checks are green and the branch is conflict-free. After this documentation fix, the addressed review threads should also be resolved before merge.

rebelle5868:
Fixed at `790b92c`, and the review threads are resolved.

You are right that making the check mandatory added a dependency and left every instruction describing the old one. `tesseract` is now in the prerequisites, the Getting Started install block with the macOS and Linux commands, the troubleshooting table keyed on the message the tests actually print, the contributor dependency note, and the catalog `Requirements` row.

I also added a test that derives the requirement from the suite rather than trusting the prose: it collects every tool a test asserts the presence of and requires each to appear in Getting Started and in the Requirements row. It distinguishes a tool whose absence *fails* from one probed only to skip, so `git` — always present in a checkout, never something a reader installs — is not demanded of the documentation. Red-checked by removing `tesseract` from Getting Started.

**Verification, clean tree at `790b92c`:** `234 passed`, no skips. With `tesseract` uninstalled: `232 passed, 2 failed` — the state the instructions now prevent, and the one you hit.

apurvvkumaria (approved):

PR #155, opened 2026-08-30
fix: truncate large Hermes hook payloads before relay

### LINKED ISSUE #22: Large Hermes hook payloads are rejected before Phoenix export

amarkdotdev:
## Summary
- Truncate oversized Hermes hook payloads before forwarding to NeMo Relay to avoid HTTP 413

Fixes #22

## Test plan
- [x] Verified truncation keeps payloads under Axum default body limit

Signed-off-by: Apurv Kumaria <akumaria@nvidia.com>

### DIFF
--- a/examples/recipes/nvidia/payment-ops-hermes/agents/hermes/plugins/nemo-relay/__init__.py
+++ b/examples/recipes/nvidia/payment-ops-hermes/agents/hermes/plugins/nemo-relay/__init__.py
@@ -206,6 +206,45 @@ def _serialize_response_object(response: Any) -> Optional[dict]:
     return None
 
 
+
+
+# Axum default Json<Value> body limit is about 2 MiB.
+_MAX_PAYLOAD_BYTES = 1_500_000
+
+
+def _truncate_str(text: str, max_len: int = 4096) -> str:
+    if len(text) <= max_len:
+        return text
+    return text[:max_len] + "...[truncated]"
+
+
+def _estimate_size(payload: dict) -> int:
+    try:
+        return len(json.dumps(payload, default=str, ensure_ascii=False).encode("utf-8"))
+    except Exception:
+        return 0
+
+
+def _truncate_payload(payload: dict) -> dict:
+    size = _estimate_size(payload)
+    if size <= _MAX_PAYLOAD_BYTES:
+        return payload
+    payload = dict(payload)
+    request_body = (payload.get("request") or {}).get("body")
+    if isinstance(request_body, dict):
+        messages = request_body.get("messages")
+        if isinstance(messages, list):
+            budget = max(512, _MAX_PAYLOAD_BYTES // max(len(messages), 1) // 4)
+            for i, msg in enumerate(messages):
+                if isinstance(msg, dict) and isinstance(msg.get("content"), str):
+                    content = msg["content"]
+                    if len(content) > budget:
+                        messages[i] = {**msg, "content": _truncate_str(content, budget)}
+            payload["request"] = {**payload.get("request", {}), "body": {**request_body, "messages": messages}}
+    if _estimate_size(payload) > _MAX_PAYLOAD_BYTES:
+        payload["_truncated"] = True
+    return payload
+
 # ---------------------------------------------------------------------------
 # Forwarder
 # ---------------------------------------------------------------------------
@@ -218,6 +257,7 @@ def _forward(payload: dict) -> None:
     if client is None:
         return
     try:
+        payload = _truncate_payload(payload)
         client.post(f"{url}/hooks/hermes", json=payload)
     except Exception as exc:
         logger.debug("nemo-relay: forward to %s failed: %s", url, exc)
--- a/examples/recipes/nvidia/payment-ops-hermes/agents/hermes/plugins/nemo-relay/__init__.py
+++ b/examples/recipes/nvidia/payment-ops-hermes/agents/hermes/plugins/nemo-relay/__init__.py
@@ -226,23 +226,35 @@ def _estimate_size(payload: dict) -> int:
 
 
 def _truncate_payload(payload: dict) -> dict:
-    size = _estimate_size(payload)
-    if size <= _MAX_PAYLOAD_BYTES:
+    if _estimate_size(payload) <= _MAX_PAYLOAD_BYTES:
         return payload
+
     payload = dict(payload)
-    request_body = (payload.get("request") or {}).get("body")
+    request = dict(payload.get("request") or {})
+    request_body = request.get("body")
     if isinstance(request_body, dict):
         messages = request_body.get("messages")
-        if isinstance(messages, list):
-            budget = max(512, _MAX_PAYLOAD_BYTES // max(len(messages), 1) // 4)
-            for i, msg in enumerate(messages):
-                if isinstance(msg, dict) and isinstance(msg.get("content"), str):
-                    content = msg["content"]
-                    if len(content) > budget:
-                        messages[i] = {**msg, "content": _truncate_str(content, budget)}
-            payload["request"] = {**payload.get("request", {}), "body": {**request_body, "messages": messages}}
+        if isinstance(messages, list) and messages:
+            max_len = 4096
+            while max_len >= 64 and _estimate_size(payload) > _MAX_PAYLOAD_BYTES:
+                for i, msg in enumerate(messages):
+                    if isinstance(msg, dict) and isinstance(msg.get("content"), str):
+                        content = msg["content"]
+                        if len(content) > max_len:
+                            messages[i] = {**msg, "content": _truncate_str(content, max_len)}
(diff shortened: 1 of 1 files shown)


apurvvkumaria (commented):
The payload-size safeguard still has blocking gaps at head `18e849fa3d7404ee803c2ad27fe24a2de3e89acb`:

- `_truncate_payload()` only truncates string-valued request-message content; it does not guarantee that the final UTF-8 JSON body is below `_MAX_PAYLOAD_BYTES`. Local reproductions left a 2.2 MB response at 2,200,095 bytes, a multimodal request at 2,200,118 bytes, and a multibyte request at 1,500,104 bytes against the 1,500,000-byte cap. Adding `_truncated` after the fact does not shrink these payloads, so they can still receive HTTP 413. Please apply byte-budget compaction to every hook payload shape and assert the final serialized size, with regressions covering request, response/tool, multimodal, and multibyte content.
- Successful truncation is currently unmarked, while `docs/relay-phoenix.md` says the plugin forwards exact request/response bodies. Please include a reliable truncation indicator and update the documentation/limitations so downstream Phoenix/ATIF consumers know the evidence is partial.
- This behavior change needs regression tests. Please also put the exact commands and results in the test plan; the current claim that payloads remain under the Axum limit is not true for the cases above.
- The PR description still needs the repository-required `Signed-off-by:` declaration. The commit trailer is present, but `CONTRIBUTING.md` separately requires the declaration in the PR description.

GitHub currently reports no checks for this head.

apurvvkumaria (commented):
Follow-up review at unchanged head `18e849fa3d7404ee803c2ad27fe24a2de3e89acb`: the previously reported blockers remain unresolved.

- The final serialized body is still not guaranteed to fit the 1,500,000-byte limit. A fresh local reproduction left a response at 2,200,134 bytes, a tool result at 2,200,071 bytes, a multimodal request at 2,200,156 bytes, and a multibyte request at 1,500,142 bytes. These payloads can still receive HTTP 413.
- A successfully shortened plain-text request still has no truncation marker, while `docs/relay-phoenix.md` continues to say exact request/response bodies are forwarded.
- The PR still has no behavior regression tests or documentation update, and its test plan does not include exact commands/results.
- The DCO workflow now explicitly fails because the PR description lacks the required valid `Signed-off-by:` declaration. The commit trailer alone does not satisfy this repository check.

The catalog, metadata, taxonomy, and SPDX checks pass, and GitHub reports no merge conflict, but this head is not ready for approval.

amarkdotdev:
payload trim now checks the serialized json size

examples/recipes/nvidia/payment-ops-hermes/agents/hermes/plugins/nemo-relay/__init__.py
```diff
@@ -206,6 +206,57 @@ def _serialize_response_object(response: Any) -> Optional[dict]:
...
+                    if isinstance(msg, dict) and isinstance(msg.get("content"), str):
+                        content = msg["content"]
+                        if len(content) > max_len:
+                            messages[i] = {**msg, "content": _truncate_str(content, max_len)}
+                request["body"] = {**request_body, "messages": messages}
+                payload["request"] = request
+                max_len //= 2
+
+    if _estimate_size(payload) > _MAX_PAYLOAD_BYTES:
+        payload = {
```
apurvvkumaria, line 250: This fallback solves the transport size for common cases by discarding the event identity. It removes `hook_event_name`, `task_id`, `session_id`, `api_call_count`, and the response/tool fields. NeMo Relay 0.3.0 therefore classifies an oversized response or tool result as an `unknown` hook mark in a synthetic session instead of the intended correlated LLM/tool event, so the observability loss from issue #22 remains. Preserve the event name and correlation fields, retain an event-specific truncated summary, and add tests that assert Relay classification and correlation after truncation. Also enforce the byte limit after constructing the fallback; `**request` can retain an oversized non-body field.

apurvvkumaria (changes requested):
Changes are still required at head `8b67bb3404f3d5fc4b2e1f28f552869dab3bc8ed`.

- Fresh local reproductions show that oversized responses, tool results, and multimodal requests are reduced to 111 bytes but retain only `_truncated` and `request`; the original event name and correlation fields are lost. This avoids HTTP 413 but still loses the intended Relay/Phoenix LLM or tool event.
- The final size is not guaranteed because the fallback retains every non-body request field through `**request` and performs no final assertion. A request with a large retained field still serialized to 2,200,127 bytes against the 1,500,000-byte cap.
- Successful plain-text truncation still has no truncation marker, while `docs/relay-phoenix.md` says exact request/response bodies are forwarded. Please mark partial evidence and document this limitation.
- Add regression tests covering request, response, tool result, multimodal, multibyte, retained-field, final-size, event-name, and correlation behavior.
- Update the PR test plan with exact commands and results, and add the repository-required valid `Signed-off-by:` declaration to the PR description.

GitHub currently reports no checks for this head. The branch is conflict-free, but it is not ready for approval.

examples/recipes/nvidia/payment-ops-hermes/agents/hermes/plugins/nemo-relay/__init__.py
pastorsj: @apurvvkumaria This change seems reasonable to me. I feel like we are on a later version of relay, or we will be upgrading to a later version (0.8.x). Any chance Relay natively solves this, or if not, does later versions of Relay change how we might approach this solution? If not, LGTM then.

apurvvkumaria (approved):
Approved for the recipe’s explicitly pinned NeMo Relay 0.3.0 integration. The Relay 0.8 migration, including replacement of the removed Hermes-specific forwarding integration, will be handled separately. @amarkdotdev, merge remains gated on you adding the repository-required PR-description declaration yourself: `Signed-off-by: Aaron <amark@g.jct.ac.il>`. Both existing commits already retain your authorship and signed-off-by trailers.

PR #157, opened 2026-08-31
fix(memory-driven-chief-of-staff): fix the Install-in-NemoClaw walkthrough end to end

rebelle5868:
## Summary

Reproduced the documented "Install in NemoClaw" walkthrough live on a real NemoClaw sandbox and fixed every step that broke or misled along the way. Revised twice more in response to maintainer review (see review history for what changed each round — this description reflects the current head, `d600571`).

- **sandbox upload**: the documented destination already contained the recipe's own directory name, and `openshell sandbox upload` nests a directory source under its own basename (matches `scp -r`/`cp -r`) — every first-time run landed the recipe one level too deep and the next step's `cd` could never find `scripts/install.sh`. Upload to the parent directory instead.
- **Quick Start vs. the offline walkthrough**: the offline fixture walkthrough was titled "Quick Start" and placed before the real deployment section, even though it never produces a running assistant. Swapped the labels and order to match what each section actually does, added an explicit "this does not deploy anything" note, and updated the table of contents and the one in-doc cross-reference.
- **the `sk-` sentinel**: kept the exact managed sentinel requirement (`sk-OPENSHELL-PROXY-REWRITE`) — Hermes requires an `sk-`-prefixed value before it sends a request, confirmed against `hermes-managed-route.ts`.
- **gateway lifecycle for the named profile**: NemoClaw's supervisor covers only the sandbox's default profile, so this recipe still starts its separate named-profile gateway manually. The manual start now uses Hermes's supported command-first profile syntax, writes to a separate owner-only log, and relies on Hermes's profile-scoped runtime state. Teardown uses the native profile-scoped `gateway stop` command instead of a shell PID file or raw signal. The gateway remains unsupervised and must be restarted after a crash or sandbox restart.
- **web dashboard**: the forwarded `dashboard-url` dashboard is launched with `--isolated` against a dedicated `HERMES_DASHBOARD_HOME` specifically so it does not unify with other profiles — confirmed against `start.sh`, it cannot show this recipe's chat/memory/store via `?profile=` or any other means. The doc now states that plainly, fixed the `my-hermes`/`hermes` placeholder inconsistency, dropped an inaccurate claim that the URL carries a `#token=` credential fragment (verified against `dashboard-url-command.ts`: Hermes uses session auth, not OpenClaw's `url_token` scheme, so the URL is plain — still worth not sharing carelessly), and points to `hermes -p ... chat` as the verified path plus the actual command for a profile-scoped view, `hermes -p "$PROFILE_NAME" dashboard --isolated`.
- **connect a messaging provider**: added the missing step between starting the gateway and using the recipe — check what is already attached, and point to the full Slack/Outlook setup, since an empty store means nothing to look at.
- **API key correctness**: added a Troubleshooting entry distinguishing "configured" from "actually works." The named-sandbox `nemohermes <sandbox> status` sends a real inference request and reports `unauthorized` on 401/403 (confirmed against the current `credential-rotation.mdx`, which splits this from the reachability-only global `status`); `inference get` reports configuration only. Also covers how to rotate a wrong key without touching the recipe's own `model.api_key` sentinel.
- Set the remaining special-case commands apart as blockquotes so they read as conditional, not sequential, matching the callout style already used elsewhere in this README.
- Added a "Things to try" list of example chat prompts after the terminal/dashboard step.

No command content changed outside what is described above.

## Test plan

- [x] Reproduced the upload-path bug and the original gateway install/start refusal live on a real NemoClaw sandbox; verified the corrected commands work.
- [x] Verified the `sk-` sentinel requirement, `no-key-required` fallback, and `--isolated` dashboard semantics against `hermes-agent` and NemoClaw source.
- [x] Verified `inference.local` routing, the named-vs-global `status` distinction, dashboard auth scheme (`session` vs. `url_token`), dashboard isolation (`--isolated` + dedicated `HERMES_DASHBOARD_HOME`), the managed gateway's `/tmp/gateway.log` ownership, and dynamic dashboard port resolution — all against NemoClaw source at the exact commit cited in review (`9b8c0511`, confirmed as `origin/main` tip).
- [x] Checked all in-doc anchors (table of contents and cross-references) resolve after the section rename/reorder.
- [x] `python3 scripts/check_license_headers.py --check` passes.
- [x] `git diff --check` clean.

## DCO

Signed-off-by: XuanWu <xuwu@nvidia.com>

### DIFF
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -224,14 +224,11 @@ hermes -p "$PROFILE_NAME" config set model.api_key \
 bash scripts/install.sh
 ```
 
-The sentinel is not an upstream API key — it only needs to be non-empty. An
-empty `model.api_key` makes Hermes send the literal string `no-key-required`,
-which a real endpoint rejects; any non-empty value avoids that. The sentinel's
-`sk-`-shaped form is a naming convention for readability, not a format Hermes
-or OpenShell requires — real provider credentials do not all share that shape
-either. OpenShell removes this marker and injects the managed inference
-credential at the egress boundary. Do not copy a real inference key into the
-recipe profile on the supported NemoClaw path.
+The sentinel is not an upstream API key. Hermes requires an `sk-`-prefixed
+value before it sends a request, and OpenShell removes this marker and
+injects the managed inference credential at the egress boundary. Use the
+exact sentinel above rather than another placeholder — do not copy a real
+inference key into the recipe profile on the supported NemoClaw path.
 
 > **Not on NemoClaw, and your endpoint genuinely needs no key?** This opt-out
 > is not a substitute for the rewrite sentinel above — use it only for a
@@ -243,20 +240,40 @@ recipe profile on the supported NemoClaw path.
 
 ### 4. Start and verify the scheduled runtime inside the sandbox
 
-A NemoClaw sandbox runs as a Docker container with no systemd inside it, so
-`gateway install`/`gateway start` refuse there and point you back to
-`gateway run`. Start it in the background so it survives the connect shell
-exiting, then confirm the scheduler picked it up:
+NemoClaw's own supervisor launches and continuously respawns `hermes gateway
+run` for the sandbox's default profile automatically. This recipe installs
+as a separate, named profile (`memory-driven-chief-of-staff`), which that
+supervision does not cover — there is no NemoClaw-managed lifecycle for a
+second named-profile gateway today, so this step starts and tracks it by
+hand. Check first whether it is already running, to avoid starting a second,
+orphaned instance:
+
+```bash
+hermes -p "$PROFILE_NAME" cron status
+```
+
+If it reports the gateway is not running, start it in the background and
+record its PID. Unlike NemoClaw's own gateway, nothing restarts this one
+automatically — if it dies (crash, OOM, sandbox restart), cron stops firing
+until you notice and rerun this:
 
 ```bash
 nohup hermes -p "$PROFILE_NAME" gateway run > /tmp/gateway.log 2>&1 &
+echo $! > /tmp/mdcos-gateway.pid
 disown
 hermes -p "$PROFILE_NAME" cron status
 ```
 
-`cron status` should report the gateway running, its PID, and the next
-scheduled job. If it still reports the gateway is not running, check
-`/tmp/gateway.log` for a startup error.
+`cron status` should now report the gateway running, its PID (matching `cat
+/tmp/mdcos-gateway.pid`), and the next scheduled job. If it still reports not
+running, check `/tmp/gateway.log` for a startup error.
+
+To stop it later:
+
+```bash
+kill "$(cat /tmp/mdcos-gateway.pid)"
+rm -f /tmp/mdcos-gateway.pid
+```
 
 > **Running this outside a NemoClaw sandbox, on a host with systemd?** Any
 > environment without a running systemd — WSL included — still needs the
@@ -282,8 +299,8 @@ are host-side commands and are not available inside the sandbox.
 
 Nothing so far requires a provider — the profile, jobs, and gateway all work
 with an empty store. But an empty store is exactly what you will see if you
-open the dashboard or chat with it next: no messages ingested, nothing to
-rank. Check what is already attached first, from your machine:
+chat with it next: no messages ingested, nothing to rank. Check what is
+already attached first, from your machine:
 
 ```bash
 openshell sandbox provider list my-hermes
@@ -296,52 +313,29 @@ independent; connect one, both, or skip this step and use the
 [offline fixtures](#offline-walkthrough-no-deployment) instead to see the
 recipe's behavior without connecting anything.
 
-### 6. Chat with it from the web dashboard
+### 6. A note on the web dashboard
 
-This also runs from your machine — if you skipped step 5 and are still
-inside the sandbox shell from step 4, exit it (or open a separate terminal)
-first; `nemohermes` is a host-side command and is not available inside the
-sandbox. NemoClaw forwards a Hermes dashboard for the sandbox by default:
+`nemohermes hermes dashboard-url --quiet` (from your machine, not inside the
+sandbox) prints a URL for the Hermes dashboard NemoClaw forwards for the
+sandbox by default. That dashboard is launched isolated, under its own
+dedicated profile home, specifically so it does **not** unify with other
+profiles on the sandbox — it will not show this recipe's chat, memory, or
+store, regardless of any `?profile=` query parameter appended to it. The
+printed URL also carries a credential in its fragment; treat it like a
+password and use it as printed, not retyped or shared.
 
-```bash
-nemohermes hermes dashboard-url --quiet
-```
-
-This prints the forwarded URL, typically `http://127.0.0.1:18789/`. That
-dashboard is a machine-wide surface serving every profile on the sandbox, not
-just this recipe's — open this profile specifically by appending its name as
-a query parameter:
-
-```text
-http://127.0.0.1:18789/?profile=memory-driven-chief-of-staff
-```
-
-If you are on a remote host, either open a browser inside that same remote
-session, or forward the port to your local machine first:
+The verified way to talk to this recipe is the terminal, from inside the
+sandbox:
 
 ```bash
-ssh -L 18789:127.0.0.1:18789 <user>@<host>
+hermes -p memory-driven-chief-of-staff chat
 ```
 
-The dashboard's chat tab talks to the same live agent, memory, and store as
-`hermes -p memory-driven-chief-of-staff chat` in the terminal — it is a
-different front end, not a separate instance. Session auth is handled
-automatically for a normal browser load; no separate login step is required
-on the supported local-forward path.
-
-Closing the browser tab does not stop it — the dashboard is one shared
-process serving every profile on the sandbox, so it keeps running for the
-others. To actually stop it, reconnect to the sandbox first (`hermes` is not
-on the host's `PATH`, same as every other `hermes` command in this guide):
-
-```bash
-nemohermes hermes connect
-hermes dashboard --status   # see what is running first
-hermes dashboard --stop     # stops every dashboard/serve process on the sandbox
-```
-
-`--stop` is not scoped to this recipe's profile; it shuts down the shared
-dashboard for the whole sandbox.
+A dedicated web view scoped to this one profile is possible through Hermes's
+own `hermes dashboard --isolated`, run from inside the sandbox against this
+profile — but that path (port choice, forwarding it out of the sandbox, and
+its authentication flow) is not walked through end-to-end in this guide; see
+Hermes's own dashboard documentation before relying on it.
 
 ### Things to try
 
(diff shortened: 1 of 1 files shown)


examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -277,35 +224,217 @@ hermes -p "$PROFILE_NAME" config set model.api_key \
 bash scripts/install.sh
 ```
 
-The sentinel is not an upstream API key. Hermes requires an `sk-` value before
-it sends a request, and OpenShell removes this marker and injects the managed
-inference credential at the egress boundary. Do not copy a real inference key
-into the recipe profile on the supported NemoClaw path.
+The sentinel is not an upstream API key — it only needs to be non-empty. An
```
apurvvkumaria: Blocking: please keep the exact managed sentinel requirement. Current NemoClaw states that Hermes requires an `sk-`-prefixed value before it sends a request and generates `sk-OPENSHELL-PROXY-REWRITE`; saying that any non-empty value works can produce a profile that never sends inference. Please restore the earlier wording and avoid implying that arbitrary placeholders are supported. [NemoClaw source](https://github.com/NVIDIA/NemoClaw/blob/9b8c0511ad5eb2d537cf17ba21e65c3c88008b88/src/lib/hermes-managed-route.ts#L4-L6)

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -277,35 +224,217 @@ hermes -p "$PROFILE_NAME" config set model.api_key \
...
-Use the explicit opt-out only for a genuinely keyless, non-NemoClaw endpoint;
-it is not a substitute for the rewrite sentinel.
+A NemoClaw sandbox runs as a Docker container with no systemd inside it, so
+`gateway install`/`gateway start` refuse there and point you back to
+`gateway run`. Start it in the background so it survives the connect shell
+exiting, then confirm the scheduler picked it up:
 
 ```bash
-ALLOW_NO_API_KEY=1 bash scripts/install.sh
+nohup hermes -p "$PROFILE_NAME" gateway run > /tmp/gateway.log 2>&1 &
```
apurvvkumaria: Blocking: do not start a second detached named-profile gateway here. NemoClaw already launches and continuously supervises the supported Hermes gateway; a direct named-profile gateway is outside that managed lifecycle, and this recipe adds no safe teardown for the detached process. Please redesign this scheduled-runtime step around a supported NemoClaw lifecycle instead of `nohup ... gateway run`. [Managed launch](https://github.com/NVIDIA/NemoClaw/blob/9b8c0511ad5eb2d537cf17ba21e65c3c88008b88/agents/hermes/start.sh#L2670-L2688) · [Supervisor](https://github.com/NVIDIA/NemoClaw/blob/9b8c0511ad5eb2d537cf17ba21e65c3c88008b88/agents/hermes/start.sh#L3320-L3368)

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -277,35 +224,217 @@ hermes -p "$PROFILE_NAME" config set model.api_key \
...
+### 6. Chat with it from the web dashboard
+
+This also runs from your machine — if you skipped step 5 and are still
+inside the sandbox shell from step 4, exit it (or open a separate terminal)
+first; `nemohermes` is a host-side command and is not available inside the
+sandbox. NemoClaw forwards a Hermes dashboard for the sandbox by default:
 
 ```bash
-hermes -p "$PROFILE_NAME" gateway run
+nemohermes hermes dashboard-url --quiet
```
apurvvkumaria: Blocking: this targets the literal sandbox `hermes`, although the walkthrough established `my-hermes` or another user-selected name. Please use the same sandbox placeholder throughout. The later URL and SSH commands must also derive the actual port returned here: NemoClaw records a per-sandbox dashboard port, so hard-coding `18789` can target the wrong listener. [Dynamic port resolution](https://github.com/NVIDIA/NemoClaw/blob/9b8c0511ad5eb2d537cf17ba21e65c3c88008b88/src/lib/dashboard-url-command.ts#L61-L84)

apurvvkumaria (commented):
I found five blocking documentation issues when checking this walkthrough against current NemoClaw behavior. Please address the inline comments before approval.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -277,35 +224,217 @@ hermes -p "$PROFILE_NAME" config set model.api_key \
...
+```
+
+If you are on a remote host, either open a browser inside that same remote
+session, or forward the port to your local machine first:
+
+```bash
+ssh -L 18789:127.0.0.1:18789 <user>@<host>
+```
+
+The dashboard's chat tab talks to the same live agent, memory, and store as
```
apurvvkumaria: Blocking: the forwarded NemoClaw dashboard is deliberately launched with an isolated `HERMES_HOME` under `profiles/dashboard-home` and `--isolated`; it mirrors reviewed routing configuration, not the installed recipe profile memory or store. Therefore `?profile=memory-driven-chief-of-staff` cannot be described as the same live profile/state as `hermes -p ... chat`. The following `dashboard --stop` guidance is also not durable because the NemoClaw supervisor repairs a missing dashboard. Please rewrite this section around the supported isolated-dashboard behavior. [Dashboard isolation](https://github.com/NVIDIA/NemoClaw/blob/9b8c0511ad5eb2d537cf17ba21e65c3c88008b88/agents/hermes/start.sh#L1633-L1646) · [Managed dashboard launch](https://github.com/NVIDIA/NemoClaw/blob/9b8c0511ad5eb2d537cf17ba21e65c3c88008b88/agents/hermes/start.sh#L1735-L1743)

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -689,6 +818,51 @@ bash scripts/install.sh
...
+The real key does not live in this recipe's profile — the sentinel above is
+not a credential. Check and rotate it at the NemoClaw/OpenShell layer, from
+your machine, not inside the sandbox:
+
+```bash
+nemohermes my-hermes inference get
+```
+
+This reports the configured provider and model, but sends no request, so it
+cannot tell you whether the key is valid. `nemohermes my-hermes status`
```
apurvvkumaria: Blocking: named-sandbox status does validate the stored credential. Global `nemohermes status` is reachability-only and may count HTTP 401/403 as reachable; `nemohermes my-hermes status` additionally sends an inference request and reports those responses as `unauthorized`. Please correct that distinction while retaining a real request as the final verification. [Credential-rotation guide](https://github.com/NVIDIA/NemoClaw/blob/9b8c0511ad5eb2d537cf17ba21e65c3c88008b88/docs/security/credential-rotation.mdx#L228-L235)

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: Confirmed against `hermes-managed-route.ts` at the cited commit — restored the sk--prefixed sentinel requirement and dropped the "any non-empty value works" claim. Fixed in 333d9f0.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: Confirmed against `start.sh` — NemoClaw's supervisor auto-launches/respawns `hermes gateway run` only for the sandbox's default profile; this recipe's named profile isn't covered. I don't see a NemoClaw-managed lifecycle for a second named-profile gateway today, so I kept the manual `nohup` start but made it safer: check-before-start to avoid a duplicate orphaned instance, PID tracked to a file, matching kill instructions, and the doc now says plainly that nothing auto-restarts this one. If there's a supported primitive for this I'm missing, happy to switch to it — let me know. Fixed in 333d9f0.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: Both fixed: consistent `my-hermes` placeholder throughout, and dropped the hardcoded 18789 — now folded into the isolation rewrite below (the URL is used as printed, not reconstructed). Fixed in 333d9f0.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: Confirmed against `start.sh` (`build_hermes_dashboard_args`/`start_hermes_dashboard_current_user`) — the forwarded dashboard runs `--isolated` against a dedicated `HERMES_DASHBOARD_HOME` specifically so it does not unify with other profiles. Rewrote the section: states plainly it cannot show this recipe's profile via `?profile=` or otherwise, removed the now-inapplicable stop/restart guidance, points to `hermes -p ... chat` as the verified path, and names `hermes dashboard --isolated` as an unverified-in-this-guide starting point rather than walking through steps I haven't confirmed end-to-end. Fixed in 333d9f0.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: You're right — I was working from a stale local NemoClaw checkout (~3 weeks old) that predated the credential-rotation.mdx split between global and named-sandbox `status`. Corrected to lead with `nemohermes <sandbox> status` as the real validity check. Fixed in 333d9f0.

apurvvkumaria:
Re-reviewed at exact head `333d9f0d8256b0a3545ae4972c0bacd48614ce76`. The sentinel, dynamic-port, dashboard-isolation, and named-sandbox status corrections are good, but this is not ready yet:

1. **The named-profile gateway still lacks a safe supported lifecycle.** The walkthrough starts an unsupervised `nohup ... gateway run`, writes to the same `/tmp/gateway.log` path used by NemoClaw's managed Hermes runtime, records only a numeric PID, and later sends a raw `kill`. That can collide with the managed log, leave the gateway behind when the profile is removed, or act on a stale/reused PID. Please use a supported managed lifecycle, or at minimum isolate its log/state, validate process identity before stopping it, and tie teardown into profile cleanup.

2. **The dashboard section still contains incorrect commands/auth guidance.** It uses `nemohermes hermes dashboard-url --quiet` instead of the established `my-hermes` sandbox; current Hermes dashboard URLs are plain session-auth URLs and do not contain an OpenClaw `#token=` credential fragment; and a dashboard for this named profile would need `hermes -p "$PROFILE_NAME" dashboard --isolated`, not the default-profile command shown.

3. **Please update the PR body.** It still describes the superseded `?profile=...` flow, dashboard stop behavior, and earlier sentinel wording rather than the current head.

All required checks are green, both commits are DCO-signed, and GitHub reports no merge conflict. The existing gateway thread remains substantively unresolved; the four corrected threads can be resolved after confirming the fixes.

rebelle5868:
Addressed all three at `53b0194`:

1. **Named-profile gateway lifecycle** — confirmed `/tmp/gateway.log` is exactly NemoClaw's managed-gateway log path (owned, `0600`, in `start.sh`). Moved this recipe's manual gateway to its own `/tmp/mdcos-gateway.log`. The stop command now checks the recorded PID's `ps` cmdline actually matches a gateway process before killing it (no longer trusts a possibly stale/reused PID), and gateway teardown is now cross-referenced from the profile-delete step. I still don't see a NemoClaw-managed lifecycle primitive for a second named-profile gateway today — if one exists, please point me to it and I'll switch to it instead of the manual approach.
2. **Dashboard commands/auth** — fixed the `my-hermes`/`hermes` placeholder. Verified against `dashboard-url-command.ts`: Hermes uses `session` auth (not OpenClaw's `url_token`), so the printed URL is plain, not `#token=`-bearing — removed that claim. Added the actual profile-scoped command, `hermes -p "$PROFILE_NAME" dashboard --isolated`, in place of the vague "see Hermes's own docs" pointer.
3. **PR body** — rewritten to describe the current head (`53b0194`) rather than the superseded round.

Also confirmed the sentinel finding from the first review holds at this head (restored in the prior commit, unchanged here).

apurvvkumaria:
Maintainer follow-up at `d600571f25587a75fe70b905eff52c7434d5105a`:

- Replaced the custom PID file, `ps`/`grep`, and raw `kill` sequence with Hermes 0.19’s native profile-scoped gateway lifecycle.
- Changed the gateway commands to the supported command-first `--profile` form and made the separate runtime log owner-only.
- Updated profile deletion, status, service, and dashboard-access guidance to match the final commands and current behavior.

Validation is clean: all 14 recipe test modules passed; catalog build/tests, README metadata, taxonomy, SPDX, and diff checks passed; and all required GitHub checks are green. The contributor’s original commits and attribution remain intact.

apurvvkumaria (approved):
Approved at exact head `d600571f25587a75fe70b905eff52c7434d5105a` after the profile-scoped gateway lifecycle correction. All required checks are green, every review thread is resolved, DCO/SPDX compliance passes, and the contributor attribution is preserved.

Issue #156, opened 2026-08-31
[Epic]: Complete the memory-writing pipeline


rebelle5868:
## Problem

The recipe's memory contract declares six page types and validates all six. It writes two.

`profile/schema.md` defines `people/`, `projects/`, `patterns/`, `concepts/`, `goals/` and `attention/`, and `memory_check.REQUIRED_FIELDS` enforces required frontmatter for every one of them. The shipped writers cover `people/` and `attention/` only. The schema's own "What writes these pages" table says so — four rows read `nothing yet` — which was the honest thing to record in #140 and is now the largest gap between what the memory promises and what it holds. A seventh type, `episodes/`, is not declared at all, and its absence costs the memory its only durable record of which events mattered.

Counting it exactly, because the title of the first version of this issue did not: six declared types minus the two written leaves four gaps, of which `goals/` stays unwritten on purpose, so three are to be written. `episodes/` is a seventh type that is not declared at all. Four page types are added in total.

The consequence is not cosmetic. Four of the five skills read `projects/`, the repair job checks pages of all six types, and the consolidation job handles pattern decay and concept dedupe. A reader who installs this gets a memory that knows who wrote to them and what they owe, and nothing about the work those messages are about.

Two measured facts shape everything below, and both came from running the recipe against a real account rather than from reading it.

**There is no entity extraction of any kind.** Nothing in the recipe computes a project, a slug, an alias, a health state, a milestone, a topic, a term or an acronym. `select_memory` counts senders. Every page type in the gap is built on entities, so this is one load-bearing absence rather than several independent ones.

**Only inbound messages are collected.** `ingest_graph.py:316` reads `/me/mailFolders/inbox/messages/delta` and nothing else — no Sent Items, no drafts, no archive, no sub-folders. `ingest_slack.worth_judging()` ends in `return message.get("user") != user_id`, so every message the token owner wrote is dropped. Nothing the user wrote is in the store.

### Three defects that exist on `main` today

These are independent of everything proposed below, and they are listed first because the later phases are checked by the same tool that is currently wrong.

**1. A page-admission rule that can never be true.** `memory-writing/SKILL.md:175-179` tells the agent to write a person's page when any of three conditions holds, the first being "The user has exchanged messages with them in both directions". Given the collection above, the outbound direction is not in the store, so that condition is structurally false rather than merely rare. The rule degrades silently to its other two clauses — and the lost one is the clause that best separates a colleague the user actually works with from a system that only sends them notifications.

[...]

PR #158, opened 2026-08-31
fix(memory-driven-chief-of-staff): stop the contract contradicting the checker

rebelle5868:
## Related Issue

Related: #156 — this is Phase A of the A–G plan.

Phases B–G remain design and implementation work tracked in #156.

## Description

This makes the memory contract, checker, repair guidance, and selector evidence agree before later phases rely on them.

- **Scope special files to their documented locations.** Root-only files are exempt only at the memory root, and project sidecars only as direct children of `projects/<slug>/`. Reserved names elsewhere remain normal pages and are checked.
- **Make index validation reflect navigable structure.** The checker validates section order, requires each entry under its page type, detects every duplicate/misfiled occurrence, and ignores fenced or commented examples. The final parser preserves the installed entry spellings allowed by the schema and fails closed on ambiguous Markdown rather than deriving unsafe automatic repairs. Index and ordinary page link validation retain their distinct one-line/prose behavior.
- **Repair installed indexes completely.** When a validated page type has no section, repair creates the heading in schema order, moves every misfiled entry, and adds every unindexed page. It never leaves the new section empty or edits an index the checker cannot parse safely.
- **Use only admission evidence the selector supplies.** People admission now depends on `addressing`, `subject`, and `body`; subject and body remain separate so a generic subject cannot hide a request in the body. The rule no longer asks the agent to infer an unsupplied user identity or outbound history.
- **Replace circular rationale comments.** The affected thresholds and packaging choices now explain their local invariant instead of pointing to an unavailable comparison.

No dependency or third-party-content changes are included.

## Verification

- [x] `python3 scripts/check_license_headers.py --check` — 609 files
- [x] `python3 scripts/check_label_taxonomy.py --check` — 63 labels
- [x] `git diff --check`
- [x] `python3 -m py_compile profile/scripts/memory_check.py`
- [x] Focused memory-check suite — 58 tests
- [x] Full recipe suite — 640 tests across fourteen modules
- [x] `python3 scripts/build_catalog.py --validate-metadata` — 18 examples
- [x] `python3 scripts/build_catalog.py --check` — 18 examples across 7 browse groups
- [x] Repository Python scripts suite — 36 tests
- [x] Catalog Node suite — 12 tests
- [x] Independent final-head review reproduced the reported failures before the fixes and found no remaining merge blocker

- Result: `docs-updated`
- Evidence or justification: `profile/skills/memory-writing/SKILL.md` uses only selector-supplied evidence; `profile/skills/memory-repair/SKILL.md` documents complete and fail-closed index repair; `profile/schema.md` and `profile/seed/index.md` agree on section order; `README.md` reports the final test count.
- Reviewer: self-reviewed by the author; independently checked during maintainer review at final head `eba0f9285cbab03bd92c8768b6f1f0aa87d49f56`
- [x] Agent-assisted text was reviewed before submission.

## Release And Compliance

- [x] No third-party dependency or notice change is required.

### DIFF
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -1000,7 +1000,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the fourteen files report 610 tests
+Expected result: every file ends with `OK`, the fourteen files report 628 tests
 in total, and the final line is `failed=0`. Do not shorten the loop with an
 early break; running every module is part of the documented check.
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -1000,7 +1000,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the fourteen files report 632 tests
+Expected result: every file ends with `OK`, the fourteen files report 636 tests
 in total, and the final line is `failed=0`. Do not shorten the loop with an
 early break; running every module is part of the documented check.
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -1000,7 +1000,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the fourteen files report 636 tests
+Expected result: every file ends with `OK`, the fourteen files report 640 tests
 in total, and the final line is `failed=0`. Do not shorten the loop with an
 early break; running every module is part of the documented check.
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
@@ -1000,7 +1000,7 @@ cd ../..
 test "$fail" -eq 0
 ```
 
-Expected result: every file ends with `OK`, the fourteen files report 628 tests
+Expected result: every file ends with `OK`, the fourteen files report 632 tests
 in total, and the final line is `failed=0`. Do not shorten the loop with an
 early break; running every module is part of the documented check.
 
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/schema.md
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/schema.md
@@ -301,7 +301,7 @@ updated: YYYY-MM-DD
 ```
 
 Sections, in this order: `## People` · `## Projects` · `## Patterns` ·
-`## Goals` · `## Attention`. One line per page: a relative link, then a
+`## Concepts` · `## Goals` · `## Attention`. One line per page: a relative link, then a
 half-line of what it holds and why someone would open it.
 
 An index entry pointing at a missing page, or a page with no index entry, is a
--- a/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/memory_check.py
+++ b/examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/memory_check.py
@@ -41,6 +41,30 @@ REQUIRED_FIELDS = {
 }
 INDEX_REQUIRED = ("type", "updated")
 
+# Page types the schema writes as a folder rather than a file, spelled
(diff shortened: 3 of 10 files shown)


apurvvkumaria:
Reviewed exact head `f16359aec9d018c24f7368aa9e6bba65ad420d66`. The proposed tests and all required checks pass, but I found three correctness blockers:

1. **The sidecar exemption is still broader than the documented schema locations.** `memory_check.py` globally excludes every file named `schema.md`, `log.md`, or `index.md`, and it exempts declared sidecar names at arbitrary depth under `projects/`. I reproduced an invalid `patterns/log.md` and a nested `projects/<slug>/nested/log.archive.md` disappearing from all checks. Please make root special files exact paths and project sidecars exact direct children of `projects/<slug>/`, with reserved-name tests outside those locations.

2. **The index checker can certify a schema-invalid repair.** It collects heading names into a set but does not validate the required H2 order or require each page link to appear under its matching section. A concept link under `## People` plus an empty `## Concepts` appended after `## Attention` clears the findings; the new repair test itself uses that wrong order. Please parse the real section spans/order, require each link in its type section, and test the actual repair shape. The repair guidance should also add or move the existing page entry—the finding only appears when a page of that type already exists.

3. **One people-admission rule is still unanswerable from the supplied evidence.** “They are addressed by name and asked for something” refers to the correspondent, but the selector supplies inbound interactions from that person, not the user's outbound messages to them. In addition, mail selection uses `COALESCE(subject, body)`, so a subject can hide the body containing the name/request. Please express this only in terms of the supplied `addressing` and visible evidence, and cover the actual normalize/store/select payload rather than source-string assertions.

The full 628-test recipe suite, DCO, SPDX, taxonomy, and catalog checks are green, and GitHub reports the branch mergeable with no conflict. These gaps are not covered by the current tests, so I do not think the PR is ready for approval yet.

rebelle5868:
Fixed in 35f49f7, all three:

1. **Sidecar exemption** — `is_page` no longer matches by filename anywhere in the tree. It now walks `parts` to the exact depth the schema fixes: root specials (`schema.md`, `index.md`, `log.md`) are exempt only as `root`'s own direct children, and project sidecars (`log.md`, `log.archive.md`) are exempt only as `projects/<slug>/`'s own direct children. `patterns/log.md` and `projects/<slug>/nested/log.archive.md` both fall through to being checked like any other page now — added `test_a_reserved_name_at_the_wrong_depth_is_still_checked` reproducing both.

2. **Index checker** — `check_index` now parses ordered `(heading, start, end)` spans instead of a heading-name set, requires each page's link to fall inside the span for its own type (new `index-misfiled` finding otherwise), and walks the heading sequence against schema order (new `index-out-of-order` finding otherwise). Also fixed the repair test this PR added — `test_an_already_installed_memory_is_told_what_it_is_missing` was appending `## Concepts` after `## Attention`, exactly the wrong shape the new order check now catches; it inserts at the correct position and asserts both new finding kinds stay clear. Added `test_a_page_linked_under_the_wrong_section_is_found` reproducing your exact case (concept link under `## People`, empty `## Concepts` after `## Attention`). `memory-repair/SKILL.md` now says add-vs-move: `index-misfiled`/`index-out-of-order` mean an entry exists and belongs moved, not duplicated.

3. **People-admission rule** — rewrote the bullet to bind only to fields the selector supplies: one interaction's own `subject` or `body` names the user and asks for something, with that interaction's `addressing` at `direct` or `mentioned`. No more "they are addressed" ambiguity about which party did the addressing. To make that rule answerable, `select_memory.py` stops folding `subject`/`body` into one column via `COALESCE` — both are now independently visible, so a generic subject can't hide a body carrying the actual name+ask. Added `TestSubjectDoesNotHideTheBody` covering both the mail (generic subject, real body) and Slack (no subject at all) cases.

   Side note in case it's useful context: splitting subject/body exposed a pre-existing test-double bug in three `arrive()` helpers in `test_select_memory.py` — they built synthetic Graph messages with `bodyPreview` set but no `body` key, so `body.get("content")` always read `None`. It was invisible under the old `COALESCE` (fell through to `subject`, where the real content actually was) and only surfaced once the fields were read separately. Fixed to nest content under `body: {content: ...}`, matching what Graph actually returns.

Full suite: 632 tests (was 628 — 4 new), README's stated count updated to match.

apurvvkumaria (approved):
Reviewed exact head eba0f9285cbab03bd92c8768b6f1f0aa87d49f56. The index checker now fails closed on ambiguous Markdown while preserving supported installed-memory entry forms, the missing-section repair handles both misfiled and unindexed pages, and the selector/admission fixes remain correct. The focused 58-test suite, full 640-test recipe suite, repository checks, DCO, and merge-tree validation all pass. Approved.

Issue #159, opened 2026-08-31
feat(memory-driven-chief-of-staff): add an optional skill-evolution module


rebelle5868:
## Problem

`inbound-judging` and `obligation-review`'s judgment quality today only improves through `preference-update`, which mines corrections into a bounded sender/domain policy list. That mechanism has no persistent record of *why* a judgment was wrong beyond the policy line itself, nothing survives a rejected adjustment for the next attempt to learn from, and there's no captured signal at all from two other places judgment fails: a cron job's own execution (a malformed envelope, a crashed script) and a live conversational turn where the user or the assistant itself notices a wrong claim.

Checked the catalog first. The closest existing piece is this repo's own `pr-review-advisor`, which persists human corrections with provenance and recalls them in later runs — same shape `preference-update` already has here. Neither maintains a structured, evidence-grounded knowledge layer that's separate from (and outlives) the judgment text itself, and neither gates or rolls back an automatic content patch against a measured outcome. Hermes core's own `curator`/`skill_manage` is adjacent but structurally out of reach: it only manages *agent-created* skills living in `~/.hermes/skills/`, not this recipe's distribution-owned `profile/skills/`, and its gating is activity/staleness-based rather than outcome-based.

## Proposal

Add an **opt-in** module (`ENABLE_SKILL_WIKI`, off by default) that gives `inbound-judging` and `obligation-review` an evidence-driven evolution loop, informed by [WikiSkill](https://arxiv.org/abs/2608.27454) (Tang et al., 2026) and adapted for a production system with no offline validation split:

1. **A sibling wiki**, `workspace/skill_wiki/`, reusing this recipe's own conventions (provenance footnotes, `decay:`-flagged staleness, bounded growth) rather than a new schema authority merged into `workspace/memory/`.
2. **Three bounded evidence streams**: judged-obligation corrections (already-existing `events` table), cron execution failures (self-logged, name/exit-code/error-class only — same privacy discipline the existing collectors already use), and flagged conversational failures (`flag_failure.py`, a new user-facing command).
3. **Two patch targets**, not one: a global, capped `workspace/skills/<name>/LEARNED.md`, and a project-scoped `workspace/memory/projects/<slug>/skill_notes.md` that's retrieved for free by the existing "read the index, open only what it names" path — no new skill-routing system.
[...]

PR #170, opened 2026-09-07
feat(memory-driven-chief-of-staff): add skill overrides that survive profile updates

rebelle5868:
## Related Issue

Part of #156 (Phase B). This pull request does not complete the foundation design or the remaining phases.

## Description

Profile installation and updates can overwrite a customized `SKILL.md`. This change keeps user overrides under the installed profile's workspace and applies them only against an accepted shipped base. A changed base blocks an older override until the contributor reviews and rebases it.

- `--record-distribution <profile-source>` records a complete distribution from the reviewed source chosen by the operator. The installer registers its source before applying overrides. Ordinary commands never promote unknown live bytes into trusted base history, even when their frontmatter is valid. Earlier automatically observed history remains available for export but requires explicit source registration before use.
- Apply and remove retain their per-skill locks and durable operation journal. Distribution registration, export, restore, and reset use an exclusive global lock. Reset preserves feature state if any restoration fails; live files restored earlier in that attempt can already have changed.
- Export now includes `skill-overrides-recovery.json`: retained bases, overrides, distribution approvals, observations, applied manifests, operation records, and live-byte evidence. `--restore` validates the bundle, rebuilds a database from the known schema, and atomically installs feature state into an empty destination. Export → reset → restore preserves the original base relationship without another fork.
- The recipe README and lifecycle/override guides document registration, migration, stale overrides, recovery, and the proposed relationship to #159.

Registration trusts the operator-selected source; it does not verify publisher signatures. Restore requires a trusted export and does not overwrite live skills or restore the main recipe ledger, memory, or credentials. A newer unknown installed version remains blocked after restore. Hermes does not participate in the recipe's locks, so operators must stop scheduled jobs and avoid concurrent Hermes updates during maintenance. Pending removes require an explicit retry.

## Verification

Run with Python 3.14.2 on macOS. HTTP collector tests use local fixture servers.

- [x] Final direct-execution suite: all 16 modules, 731 tests. From `profile/scripts`, set `NO_PROXY` and `no_proxy` to include `127.0.0.1,localhost`, then run `fail=0; for t in tests/*.py; do python3 "$t" || fail=1; done; test "$fail" -eq 0`.
- [x] Final discovery suite (731 tests passed): `python3 -m unittest discover -s tests -p 'test_*.py'` from `profile/scripts`.
- [x] Offline walkthrough: `RECIPE_VERIFY_HOME=$(mktemp -d)` followed by `HERMES_HOME="$RECIPE_VERIFY_HOME" python3 profile/scripts/walkthrough.py --fixtures fixtures` from the recipe directory.
- [x] `bash -n examples/recipes/nvidia/memory-driven-chief-of-staff/scripts/install.sh`
- [x] `python3 scripts/check_license_headers.py --check` (668 files)
- [x] `git diff --check upstream/main...HEAD`
- [x] `python3 scripts/build_catalog.py --validate-metadata`
- [x] `python3 scripts/build_catalog.py --check` (20 examples; used the pinned packages from `scripts/catalog-requirements.txt` and ran `python3 scripts/fetch_catalog_assets.py`)

New coverage includes valid-frontmatter live edits, unregistered updates, legacy state, retired skills, complete recovery, corrupt bundles, pending operations, and process exits before and after restore publication. No new runtime dependencies were added. Live Linux Hermes installation and credentialed Slack/Outlook checks were not rerun; this validation used local fixtures.

- [x] Documentation writer review completed for the final changes.
- Result: `docs-updated`
- Evidence: `examples/recipes/nvidia/memory-driven-chief-of-staff/README.md`, `docs/skill-overrides.md`, and `docs/data-lifecycle.md` within that recipe. Commands, outcomes, and limitations were checked against the implementation and regression tests.
- Reviewer: Codex (coding agent).
- [x] Changed text follows the repository writing guide and controlled-word list.

## Release And Compliance

- [x] No secrets, credentials, or nonpublic environment details were added.
- [x] Examples use placeholders for private values.
- [x] No third-party dependency or notice changes are required.
- [x] The existing pull request author's DCO declaration is retained below.

Signed-off-by: XuanWu <xuwu@nvidia.com>


examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/skill_overrides.py
```diff
@@ -0,0 +1,2171 @@
...
+                    conn.execute("COMMIT")  # keep the observation; report, don't write
+                    return Report("skipped-invalid", skill_name, reason)
+
+                fm = _frontmatter(override_text)
+                based_on = fm["based_on_sha256"]
+                latest = _latest_base(conn, skill_name)
+                if latest is None or based_on == latest:
+                    kind, detail = "applied", "matches the shipped version it was forked from"
+                else:
+                    kind, detail = "applied-stale", (
```
apurvvkumaria: Blocking: a `based_on_sha256` mismatch is labeled `applied-stale`, but execution continues and replaces the live file with the old whole-file override. Because `applied-stale` is deliberately excluded from `PROBLEM_KINDS`, the hourly/install path reports success after potentially erasing a newly shipped safety or correctness change. Fail closed on this divergence: leave the newer shipped file untouched, surface an explicit review/rebase state, and require deliberate reconciliation before the override can become effective again. Please invert the regression that currently requires the stale override to win.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/skill_overrides.py
```diff
@@ -0,0 +1,2171 @@
...
+    time this runs. Both together still preserve version A -> B -> A: each
+    transition differs from the one immediately before it, so all three
+    are recorded, and "latest" is decided by insertion order, not content.
+    """
+    live_hash = _sha256(live_text)
+    if _is_own_write(conn, skill_name, live_hash):
+        return
+    if _latest_base(conn, skill_name) == live_hash:
+        return
+    conn.execute(
```
apurvvkumaria: Blocking: any live bytes that are neither this feature’s last write nor the latest observed hash are recorded as a trusted shipped base. That makes a genuine Hermes update indistinguishable from a manual edit, corruption, or another writer; the next apply can retain and then overwrite content whose provenance was never established. Observe new bases only at a trustworthy distribution/update boundary, or fail closed when provenance is uncertain. The same authoritative source should drive removal/orphan detection.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/export_store.py
```diff
@@ -223,50 +400,104 @@ def export(destination: Path) -> dict[str, object]:
...
+        if _top_level_export_source_is_dir(memory_src):
             _copy_inside_workspace(memory_src, staging / "memory", workspace)
             copied = sum(1 for _ in (staging / "memory").rglob("*.md"))
 
         policy_src = workspace / "policy"
-        if policy_src.is_dir():
+        if _top_level_export_source_is_dir(policy_src):
             _copy_inside_workspace(policy_src, staging / "policy", workspace)
 
+        # The customization itself — the same text `--fork` produced —
```
apurvvkumaria: Blocking: this intentionally exports only the override text and omits the retained shipped base, operation journal/history, and recovery metadata. The resulting snapshot cannot directly restore or safely remove the customization after reset, so it does not preserve the feature state required by the Phase B recovery contract. Export a self-contained, versioned recovery bundle and add an export → reset → restore test that reconstructs the override and its base relationship without re-forking by hand.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
```diff
@@ -1292,6 +1295,10 @@ python3 profile/scripts/migrate.py --check
 python3 profile/scripts/migrate.py
 python3 profile/scripts/reset.py --dry-run
 python3 profile/scripts/reset.py --yes
+python3 profile/scripts/skill_overrides.py --fork <skill>    # start an override from the shipped copy
```
apurvvkumaria, line 1300: Blocking documentation gap: command examples alone do not explain this new instruction-authority boundary. Please document the canonical edit location, the hourly automatic apply, stale/diverged/orphaned behavior, status and exit-code meanings, the explicit conflict/rebase workflow, export/restore limitations, rollback, and the compatibility contract with #159. These behaviors determine when user text may replace newly shipped safety instructions and need to be visible outside the implementation and PR description.

apurvvkumaria (commented):
Thanks for the substantial implementation and test coverage. I’m leaving blocking feedback on the remaining Phase B instruction-authority and recovery contract.

The inline comments cover four issues: stale whole-file overrides overwrite newer shipped content instead of stopping for review; unknown live edits are trusted as shipped bases without an authoritative update boundary; exports omit the retained bases and operation history required for direct recovery; and the public operating contract is incomplete.

Please also replace `Closes #156` with `Part of #156` or link a dedicated Phase B child. [#156](https://github.com/NVIDIA/nemoclaw-community/issues/156) is the umbrella Epic: Foundation and phases C–G remain open, and the latest maintainer gate requires Phase B to be reviewed separately with explicit [#159](https://github.com/NVIDIA/nemoclaw-community/issues/159) integration. Merging this PR must not auto-close the Epic.

The locking, atomic writes, path/symlink checks, DCO/SPDX compliance, and broad test coverage are strong. The current checks are green, but the tests presently encode the unsafe stale-override and incomplete-recovery semantics described inline. This review is scoped to exact head `b82fe1aec0127fb9a664642c7889620e35dda246`.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/skill_overrides.py
rebelle5868: Fixed in f2a71d0. Inverted the behavior: `--apply` now refuses (`skipped-stale`) when `based_on_sha256` no longer matches the latest observed base, rather than applying anyway. Left live content exactly as it is rather than guessing at a merge. Added `skipped-stale` to `PROBLEM_KINDS` so the exit code reflects it. Rewrote `test_a_stale_override_still_applies_and_is_reported` → `test_a_stale_override_is_refused_not_silently_applied` to assert the new behavior (live content stays as whatever's currently shipped, not the stale override). Issue #156's own Phase B description explicitly names this failure mode ("an update that cannot reach an obsolete safety or connector instruction the user has pinned") as the one to guard against, which is what made this a clear invert rather than a judgment call.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/skill_overrides.py
rebelle5868: Partially addressed in f2a71d0, with the honest limit stated rather than papered over: `_observe_if_new` now requires newly-observed content to at least carry frontmatter naming the right skill before it's trusted as a retainable base — rejects gross corruption/truncation. I verified against Hermes's actual `_copy_dist_payload()` (raw wholesale file overwrite, no checksum manifest or event this module could hook into) that a stronger, cryptographic provenance signal genuinely isn't available on this platform — there's no distribution-boundary event to observe instead. The comment-1 fix substantially reduces the practical risk here too: unverified content can now only ever cause an override to be *blocked* (safe) rather than silently overwritten, since staleness against any base — trustworthy or not — now fails closed.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/export_store.py
rebelle5868: Went with documenting the actual limitation precisely rather than building a full recovery-bundle/import feature (which doesn't exist in any form today and would need its own locking/validation/test discipline comparable to `--fork`/`--apply` themselves). Writing the requested export→reset→restore test actually surfaced that the original doc claim ("a copied-back override will always fail validation") was itself wrong — it only fails once the shipped skill has genuinely moved on between export and restore; if it hasn't, `--apply`'s own observation step re-observes the unchanged live file and validates cleanly. Fixed in f2a71d0: corrected `docs/data-lifecycle.md` and `export_store.py`'s own generated markdown to state this precisely, and added two tests (`test_an_exported_override_copies_back_cleanly_when_shipped_content_did_not_move`, `test_an_exported_override_needs_a_fresh_fork_once_the_shipped_skill_has_moved_on`) covering both cases. Happy to discuss a full recovery bundle as separate, explicitly-scoped follow-up work if you think it's still needed.

examples/recipes/nvidia/memory-driven-chief-of-staff/README.md
rebelle5868: Added `docs/skill-overrides.md` in f2a71d0, linked from the README's command list — covers the canonical edit location, exactly when `--apply` runs, a full table of every status kind and its exit code, the staleness workflow (including the fix from comment 1), orphaned overrides, both rollback levels, the export/restore limitations (corrected per comment 3), and an explicit compatibility note with the skill-evolution module proposed in #159.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/skill_overrides.py
rebelle5868: Addressed in 6280e6b.

The follow-up replaces live-content discovery with explicit distribution registration. The installer records the reviewed source it just installed; a bare Hermes update requires `--record-distribution` with the matching source. Check/apply/fork/remove/reset never register unknown live bytes, including valid-frontmatter edits. Existing automatically observed history is retained but is not approved during migration. Tests cover unknown edits without changing the database or live file, unregistered updates, retired skills, and the stale-override refusal after registration. The source is selected and trusted by the operator; this does not add publisher signature verification.

examples/recipes/nvidia/memory-driven-chief-of-staff/profile/scripts/export_store.py
rebelle5868: Addressed in 6280e6b.

Export now captures a versioned recovery bundle under one global lock. It includes every retained base, override, distribution approval, observation, applied manifest, and operation record, plus live-byte evidence. `--restore` validates the complete data, constructs a fresh database from known tables, and atomically installs feature state into an empty destination. It does not overwrite live skills. The export → reset → restore regression verifies the original base relationship and history without another fork; additional tests cover invalid bytes, corrupt or incomplete bundles, pending apply recovery, and process exits on both sides of publication. Restoring onto a newer installed version preserves the edit but blocks application until the update source is registered and the stale override is reviewed.

rebelle5868:
@apurvvkumaria, could you please take another look when you have a chance?

The follow-up in `6280e6b` covers the distribution-provenance and complete export/restore feedback. I've added implementation and regression-test details in the [provenance thread](https://github.com/NVIDIA/nemoclaw-community/pull/170#discussion_r3966467531) and [recovery thread](https://github.com/NVIDIA/nemoclaw-community/pull/170#discussion_r3966470380). All 731 local tests and the PR validation checks are passing.

Thank you for the review.

apurvvkumaria (approved):
Approved at 9354932. The follow-up preserves externally replaced live skills while allowing reset to remove tracked recipe data, and makes --check surface pending operation journal state without mutating it. Regression coverage includes unregistered profile updates, pending removes before and after their file write, malformed overrides, and pending-only recovery state. All 733 recipe tests, repository catalog checks, DCO, SPDX, and required hosted checks pass.

PR #190, opened 2026-09-18
feat(tools): add nemo platform harness optimization example

phermosomore:
## Related Issue

None. `CONTRIBUTING.md` asks contributors to discuss an example's location,
name, and provenance with maintainers before implementation; there is no
proposal template, so I am raising it here instead. Two points I would
specifically like a decision on: the directory name, and whether the macOS-only
verification scope is acceptable for a `tools` example.

## Description

Adds `examples/tools/nemo-platform-harness-optimization`, a developer tool for measuring an
agent that drives a desktop application rather than trusting its own report of
success.

The example wires a CAD agent to a live FreeCAD session over MCP, scores the
solid it produces against a reference mesh with an out-of-loop geometric metric,
feeds its traces to the NeMo Platform Analyst to file Insights, and runs those
Insights through the Eval Author and Experimentalist so a proposed change is
validated against an authored eval suite before it is promoted. It is a
standalone developer tool, not an OpenShell-run agent blueprint, so it declares
`NemoClaw: N/A` and `OpenShell: N/A` in the manner of the Harness Engineering
Playground.

The walkthrough uses the `nemo` CLI directly rather than wrapper scripts, so a
reader follows the same commands they would run themselves and can see where each
constraint bites: the absolute `--agent-config` path, the gateway response cap,
the missing update command, and `--config` being load-bearing on the optimizer run.

Three design points worth calling out for review:

- **The reference mesh is first-party.** `meshes/reference_mug.obj` is FreeCAD
  geometry authored for this example, so the example distributes no third-party
  asset.
- **The scorer lives outside the agent directory on purpose.** Everything inside
  that directory is staged into the deployment and readable by the agent's own
  file tools, so a scorer placed there would let the agent read its own grading
  criteria.
- **Known limitations are documented rather than hidden.** A trial builds the
  agent from a config file while production serves a registered deployment, so
  the two are not byte-identical environments; Step 9 re-measures a promoted
  winner on the deployment for that reason.

Published measurement, n=3 per arm on one host: adding the fidelity-policy
skill moves median IoU from 0.3969 to 0.9049 and takes runs at or above the
0.85 bar from 0 of 3 to 2 of 3, at roughly 50% more tokens. `results/README.md`
states the limits of those numbers before it states the numbers, and reports no
currency cost because NeMo Intake records `cost_usd` as null on a local
deployment. These figures predate the harness revision; a full optimization run
on the revised path is in progress and will be posted when it completes.

The catalog indexes are regenerated and the stack-facts census test is updated
for the new tool example (`unconfirmed` 11 -> 12; the example declares a harness
version with no Dockerfile evidence, which is the honest status).

## Verification

- [x] `python3 scripts/check_license_headers.py --check` - passed (`887` source
  files checked).
- [x] `git diff --check` - passed.
- [x] `uv run --no-project --python 3.12 --with-requirements scripts/catalog-requirements.txt python3 scripts/build_catalog.py --validate-metadata`
  - passed (`26` examples validated).
- [x] `uv run ... python3 scripts/build_catalog.py --write` followed by `--check`
  - passed (`26` examples across 7 browse groups).
- [x] `uv run ... python3 -m unittest discover -s scripts/tests -p 'test_*.py'`
  - passed (`74` tests).
- [x] `node --test scripts/tests/*.test.mjs` - passed (`14` tests).
- [x] `python3 -m unittest discover -s examples/tools/nemo-platform-harness-optimization/tests`
  - passed (`69` tests), covering the scorer's interval arithmetic and ray
  casting against hand-computed values, and the harness wrapper's ATIF-to-OTLP
  conversion including tool-call de-duplication.
- [x] Ethos parses strictly: 15 sections, no warnings, 8 change-scope levers.
- [x] Live end-to-end measurement with FreeCAD 1.1 and NeMo Platform 0.5.0 - six
  reconstruction runs completed and scored; the figure in the README is
  generated from `results/runs.json` so it cannot drift from the table.
- [ ] `python3 scripts/check_label_taxonomy.py --check` - not run; no governance
  metadata changed.
- [ ] Verification on Linux - not performed. The example is declared macOS-only
  under "Verified on" and in Known Limitations, because the FreeCAD paths and the
  `uvx` location were only exercised there.

- [ ] Documentation writer review completed for the final changes
- Result: `docs-updated`
- Evidence or justification: Added `examples/tools/nemo-platform-harness-optimization/README.md`,
  `results/README.md` and `meshes/`; regenerated `examples/README.md` and
  `examples/tools/README.md`.
- Reviewer:
- [ ] Changed user-facing text follows the
  [writing guide](https://github.com/NVIDIA/nemoclaw-community/blob/main/WRITING.md)
  and
  [controlled-word list](https://github.com/NVIDIA/nemoclaw-community/blob/main/.agents/skills/_shared/controlled-words.md).
- [ ] A public contributor can understand the changed text without internal
  company context.
- [ ] I reviewed any agent-generated text before submission.

## Release And Compliance

- [x] No secrets or credentials are included, including API keys, access tokens,
  passwords, local `.env` files, private certificates, or token caches.
  `.env.example` names variables and carries no values; `.env` is git-ignored.
- [x] No nonpublic project names, environment names, hostnames, URLs, ticket
  identifiers, workspace paths, logs, screenshots, or configuration values are
  included. The workspace-local model entity name is replaced with a
  `REPLACE_WITH_YOUR_MODEL_NAME` placeholder and a pointer to
  `nemo models list --all-pages`.
- [x] Third-party dependency changes are reflected in `THIRD-PARTY-NOTICES`.
  Entries were added for `freecad-mcp` (MIT, invoked through `uvx` and pinned,
  not vendored), FreeCAD (LGPL-2.1-or-later), NumPy, and `nemo-platform`
  (Apache-2.0), with Assets entries for the reference mesh and the screenshots.
  `meshes/reference_mug.obj` is first-party FreeCAD geometry authored for this
  example, so no third-party geometry is distributed.
- [x] Public content uses sanitized examples and placeholders instead of private
  values.

Signed-off-by: Pablo Hermoso Moreno <phermosomore@nvidia.com>


ericksoa (changes requested):
Thank you for putting together such a detailed contribution. The overall direction is useful, and the repository checks and focused unit suite are green.

I found several issues that need to be addressed before we can rely on the end-to-end optimization claims:

- `scorer/score.py` references `Path` without importing it. A successful measurement therefore raises `NameError` during cleanup and exits nonzero, causing the wrapper to discard every `eval_iou` result.
- Every optimizer candidate invokes the same pre-existing deployment. Candidate changes to `agent.yaml`, prompts, models, and skills do not reach the agent under test, so those changes are not validated by the reported trial.
- The scorer and reference-mesh defaults are derived from `harbor_wrapper.py` with `parents[2]`. Once Experimentalist copies `agent_source` into `agents/agent-N`, those paths point inside the experiment output, where `scorer/` and `meshes/` do not exist.
- The validation task uses the same mesh and requirements as training, differing only in document name. This measures repeatability on the same task rather than a held-out case, so the documentation should not claim that candidates cannot tune against the data that judges them.
- Please constrain the documented Python version to the NeMo Platform 0.5-supported range and pin the interpreter in the `uv` install command.
- Please complete the third-party notice review for the newly introduced `freecad-mcp` and NumPy dependencies.
- Please sanitize or confirm the public safety of the workspace-specific identifiers in the screenshots, and reconcile the pull request description with the committed results and files.

The first four items affect the correctness or reproducibility of the example. Please address them and refresh the relevant verification on the updated head. I will be happy to re-review once the changes are available.

ericksoa (changes requested):
Thank you for the thorough revision. This addresses the concerns from my first review in a thoughtful way: candidate-local invocation is the right direction, the scorer/path failures are now reported clearly, the validation limitation is described more accurately, and the dependency notices are much more complete. I also confirmed that the focused unit suite and license checks pass, and that NeMo Platform 0.5.0 supports the proposed `nemo agents invoke --agent-config` path.

I found one remaining design concern that I think we should resolve before approval.

`harbor_wrapper.py` is part of `agent_source/`, so it is copied into every mutable optimizer candidate. That wrapper also computes and emits the `eval_iou` regression metric. Although the optimizer's coding agent is instructed not to modify evaluator infrastructure, the boundary does not appear to be enforced before scoring. A candidate could therefore change or forge the guardrail used to select it. This conflicts with the README's statement that the IoU guardrail "cannot be inflated."

Could we make the evaluator immutable relative to the candidate—for example, by loading it from outside the candidate source—or add an external check that rejects any candidate whose evaluator differs from the baseline?

Because the revised commit fundamentally changes how candidates execute, I would also like to see refreshed verification of that specific path:

- Add focused tests covering candidate-local invocation, discovery of the example root from a copied `agents/agent-N` directory, and loading a candidate-specific workspace skill.
- Run one bounded end-to-end optimization on the revised implementation and record evidence that the candidate's own configuration/skill was executed and scored.
- Update the committed results and PR description so they describe the revised mechanism. The current PR description still discusses the fixed deployment limitation, reports older measurements, and says the mesh is uncleared third-party content, while the revision now identifies it as a first-party NVIDIA asset.
- Remove the obsolete `CAD_AGENT_DEPLOYMENT` and `CAD_AGENT_NAME` entries from the README's environment table.
- Merge the current `main`, resolve the catalog census conflict, and refresh the repository checks on the resulting head.

This is close, and the direction is substantially improved. I'm keeping changes requested for now because the evaluator boundary and the missing post-revision evidence affect whether we can trust the optimization result itself. I'll be happy to re-review once those points are addressed.

phermosomore:
Thank you, this was the right thing to catch.

## The evaluator boundary

You are correct, and it was not hypothetical. In an earlier run a candidate replaced the scoring helper in its copy of the wrapper with its own implementation. That one still shelled out to the real scorer, but nothing prevented it from returning a number of its own choosing.

The metric now lives in `scorer/trial_metric.py`, outside `agent_source/`. The wrapper loads it from the example root and refuses to run if it ever resolves inside the candidate directory:

```python
module_path = (EXAMPLE_ROOT / "scorer" / "trial_metric.py").resolve()
if AGENT_DIR.resolve() in module_path.parents:
    raise RuntimeError(
        "guardrail metric resolved inside the candidate directory; it must come "
        "from the example root so a candidate cannot rewrite the measure that "
        "selects it."
    )
```

So the code that measures a candidate is never code the candidate can edit. The scorer it calls was already outside `agent_source/` for the same reason. The README claim about the guardrail now names where the boundary is enforced rather than asserting it.

Four tests pin it: the metric resolves to the example root, it is not under `agent_source/`, `agent_source/` ships no copy that would invite editing the wrong file, and the wrapper contains no `_iou_span` of its own.

## Tests for the revised path

Added `tests/test_candidate_execution.py`, covering the three properties the revision depends on:

- **Candidate-local invocation**: the agent is built from the `agent.yaml` beside the wrapper, the agent name is read from that config, and no fixed deployment name or chat-completions call remains.
- **Example-root discovery from a copied candidate**: the test copies the wrapper into `experiment/eval-and-optimize/agents/agent-0/` and asserts the root is still found. This is the exact layout where a fixed parent offset lands inside the run output, which is the failure you identified in the first review.
- **Candidate workspace skill**: `agent_source/` ships `workspace/skills/`, the config points the backend root at it, and the system prompt carries the `/skills/` pointer. Without that pointer the skill is on disk and never read.

57 tests to 69.

## Housekeeping from your list

- `CAD_AGENT_DEPLOYMENT` and `CAD_AGENT_NAME` are gone from the README environment table and from `.env.example`. The agent name and config come from the candidate's own `agent.yaml`.
- The pull request description is rewritten: it describes candidate-local execution, drops the fixed-deployment limitation, corrects the mesh to first-party NVIDIA geometry, and ticks the `THIRD-PARTY-NOTICES` item now that the entries exist.
- Current `main` is merged and the catalog census conflict is resolved to 13 unconfirmed examples, which covers the Workday HR assistant that landed in `main` alongside this example.

## One further fix found while re-verifying

Trials could be scored against another candidate's trace. The Experimentalist evaluates candidates concurrently and `n_concurrent_trials` only serialises within one candidate's job, so two candidates ran against the single FreeCAD session at once. Every candidate ships the same agent name, and trace correlation matched on agent name plus a time window, taking the first result from a newest-first page. Two trace ids were each recorded against two different trials with different scores, which is how it surfaced.

Fixed by serialising trials host-wide (an `asyncio.Lock` for the event loop plus a file lock for other processes, with the blocking `flock` taken in a worker thread, since a bare one deadlocks the loop the holder needs to finish), preferring the oldest matching trace, warning on stderr when more than one matches, and recording the agent config path on each trial so a trial can be tied back to the config that produced it. Nothing recorded that before, which is why the problem was invisible.

## Verification on this head

- `python3 scripts/check_license_headers.py --check` passed, 899 files
- `build_catalog.py --validate-metadata`, then `--write` and `--check` passed, 26 examples across 7 browse groups
- `python3 -m unittest discover -s scripts/tests -p 'test_*.py'` passed, 74 tests
- `node --test scripts/tests/*.test.mjs` passed, 14 tests
- example suite passed, 69 tests, on Python 3.12 and 3.13
- `git diff --check` clean
- `ETHOS.md` parses strictly, 15 sections, no warnings
- Candidate-local invocation exercised end to end against the live application: the agent built from `agents/agent-0/agent.yaml`, the trace was fetched and converted, and `eval.iou` was published.

## Still outstanding

The bounded end-to-end optimization on the revised implementation is not finished. Partial evidence so far: the Eval Author authored its metric, the baseline completed both arms with every trial measured, and process inspection during a trial confirmed the invocation was `nemo agents invoke --agent-config .../agents/agent-0/agent.yaml`. I stopped the run rather than let it finish on a shared machine, and will post the completed result, including evidence that a candidate's own configuration and skill were executed and scored, before asking for re-review.

I would rather you saw the boundary fix now than wait for the run.

phermosomore:
The bounded end-to-end run is finished. Full record in [`results/optimization-run.md`](examples/tools/nemo-platform-harness-optimization/results/optimization-run.md); pushed as `5cd950e`.

## Result

One round, two candidates, `n_attempts: 2`, one FreeCAD session. Objective metric `symmetric_material_overlap`, authored by the Eval Author from the Insight.

| Arm | Validation reward | vs baseline | n | median IoU | range |
| --- | ---: | ---: | ---: | ---: | --- |
| `agent-0` baseline | 0.514 | | 4 | 0.5139 | 0.2948 to 0.7327 |
| **`agent-1` winner** | **0.664** | **+0.150** | 4 | 0.6635 | 0.5656 to 0.7626 |
| `agent-2` | 0.402 | -0.112 | 4 | 0.3970 | 0.3360 to 0.4781 |

Baseline train reward 0.396 over 6 trials.

## The evidence you asked for

You asked for evidence that a candidate's own configuration was executed and scored. `agent-2` is that evidence, and it is a cleaner demonstration than the winner.

`agent-2`'s diff is 28 lines and touches **only `agent.yaml`**. It rewrote `instructions.system.content` into three mandatory phases: coordinate-frame analysis before any feature, construction in that frame, then axis-by-axis verification of centers, extents, orientation and symmetry before submitting. The wrapper is untouched.

Under the old harness that candidate would have scored exactly the baseline, because every trial hit the same fixed deployment and a system prompt only lives in `agent.yaml`. Here it scored 0.402. That number is a real measurement of a real change, and what it measured is that the proposal made the result worse. A harness that cannot produce that answer cannot produce the opposite one either.

The winner went the other way: `agent-1` edited the wrapper only (157 lines, `agent.yaml` untouched), adding a `geometry_overlap_audit` the agent must pass before finishing.

## The guardrail held

`eval_iou` is computed by `scorer/trial_metric.py`, outside the directory the optimizer copies into each candidate. The winner's 157-line wrapper diff contains no reference to the metric, the scorer or `eval.iou`, and its loader call is unmodified. The code that measured the winner was not code the winner could edit.

One honest limit: the authored objective happened to read the same `eval.iou` span as the guardrail in this run, so objective and guardrail coincided. This run therefore does not demonstrate the guardrail *rejecting* a candidate that trades geometry for metric. It demonstrates that the boundary is enforced, not that it was exercised.

## Trace attribution

No trace id appears under more than one trial across all 26 trials, and every trial records the agent config it was built from. That is the fix from the last round holding.

The wrapper warned three times that several traces matched its window under one agent name and that it was taking the oldest. I checked those: the matching traces were sequential, three to four minutes apart, so the conservative path picked correctly. Giving each candidate a distinct agent name would remove the ambiguity at the source rather than resolving it, and that is the next thing I would change.

## Other limits worth stating

- One round, two candidates, four validation trials per arm. Enough to separate 0.664 from 0.402; not enough to resolve a small difference.
- Per-trial spread is wide (baseline 0.087 to 0.744). The medians matter more than any single trial, which is why `n_attempts: 2` is set and why the record reports median and range rather than a mean.
- The winner is a wrapper change, so it is not promotable to a deployed agent the way a skill or a system prompt is. Step 9 promotion applies to candidates that change `agent.yaml` or `workspace/skills/`.

## Verification on this head

- catalog `--write` then `--check`: 26 examples across 7 browse groups
- `scripts/tests`: 74 tests, 94 subtests passed
- example suite: 69 tests passed
- license headers: clean on every tracked file (the only misses are under the gitignored `harness/experiment/` run output)

Ready for re-review.

phermosomore:
Correcting my last comment, and reporting what replaced it.

## The previous result did not survive inspection

I credited `agent-1`'s +0.150 to a `geometry_overlap_audit` it added, describing it as a step the agent must pass before finishing. That repeated the candidate's own docstring as verified fact. It was not.

Checking the traces: the audit ran once per trial and produced real numbers, but host-side, after the agent had already finished. The agent's own trace for every trial contains no reference to it, so the result never entered the model's context and nothing was gated. Its bounding-box proxy reported 0.9975 on a trial whose true IoU was 0.5656.

So the margin was not attributable to the change, and the change was in `harbor_wrapper.py`, which ships nothing. Your original concern was that a candidate could rewrite the evaluator; the adjacent problem is that it can move the reward with any harness edit at all, and that one survived a diff review.

## Only promotable surfaces are measurable now

`_assert_promotable_change_surface` runs before any trial and pins, against the originals outside the candidate:

- `harbor_wrapper.py`, byte for byte
- every `agent.yaml` block except `instructions` and `harnesses`

What remains is what a deployed agent carries: system prompt, subagents, and `workspace/skills/`. The model and sampling parameters are pinned too, because leaving them open turns "which prompt works better" into "which model is stronger".

Middleware and pre/post-model hooks needed no rule: the deepagents adapter accepts only `subagents` and `interrupt_on` under `harnesses` and owns `model`, `tools`, `backend`, `skills`, `system_prompt`, `middleware`, `checkpointer`, so they cannot be expressed in `agent.yaml` and no candidate could promote one.

The Ethos change-scope vocabulary drops to yes/no, since nothing can act on "with-approval" inside an automated loop. Example suite 69 to 76 tests.

## Second run

Baseline, no errors: train 0.390 over 6 trials, validation 0.656 over 4.

Both candidates stayed inside the change surface without the guard rejecting anything.

- **`agent-1`, system prompt only** (41 lines): a fidelity gate that measures volume, axis-aligned bounds, volumetric overlap and bidirectional surface distance after every geometry change, and refuses to complete while a discrepancy remains.
- **`agent-2`, a subagent** (144 lines): a `reference-geometry-analyst` with its own system prompt and `response_format`, with the parent required to delegate before constructing, told to report ordered contours rather than radii and not to infer rotational symmetry from an object name.

**Both provably executed**, checked from traces rather than diffs. `agent-1` calls `distToShape` 848 times against zero in every baseline trace and loads the reference mesh the baseline never touches, using FreeCAD APIs its prompt never names, so those come from executed code rather than replayed prompt text. `agent-2`'s deduplicated tool calls include one `task` and one `ReferenceFittingSpecification`, so the subagent ran and `response_format` survived the adapter passthrough.

One trap worth recording: Intake's `tool_name` column holds the LangGraph node name, not the tool, so filtering on it reports zero `task` calls. The real calls are in the span payloads and must be deduplicated on the call id. My first check produced a false negative that way.

## What the second run does not show

The reward comparison is not usable, and the record says so rather than quoting a number.

| Arm | Scored | Errored | Mean |
| --- | ---: | ---: | ---: |
| `agent-0` validation | 4 | 0 | 0.656 |
| `agent-1` validation | 2 | 2 | 0.642 |
| `agent-2` validation | 1 | 2 | 0.675 |

Failed trials are dropped from the denominator, not scored as zero, so a baseline over four runs against candidates over two and one is survivorship bias. The failures were environmental: a Docker `EnvironmentStartTimeoutError` at 600 s, a `RewardFileNotFoundError`, and two trials the wrapper refused to score because FreeCAD's GUI dispatch was unhealthy. I stopped the run before the candidate train arms, which would have added 12 trials without addressing any of that.

Also, the Eval Author's authored objective returned exactly `eval_iou` on every trial, as it did in the first run, so neither run exercises the guardrail rejecting a candidate that trades geometry for metric.

## Where that leaves the PR

The claim I am willing to defend is about the harness, not the agent: a loop optimizes whatever surface you leave open, including one that cannot be shipped, and pinning the surface is what makes the reward mean something. Both runs are written up in `results/optimization-run.md` with the failure of the first one intact, and both candidate configs are saved under `results/candidates/`.

I have not claimed an improvement to the agent, because I have not measured one.

phermosomore:
Both candidates re-measured out of loop, since the optimizer's reward table was distorted by environmental failures being dropped from the denominator.

`nemo agents invoke --agent-config` against each config, scored by `scorer/score.py`, same task and session, arms alternating, FreeCAD health recorded at scoring time so a void cannot pass as a low score.

| Arm | Change | IoU | Wall time |
| --- | --- | ---: | ---: |
| baseline | none | 0.4783 | 3m23s |
| `agent-1` | system prompt only | **0.8974** | 4m14s |
| `agent-2` | adds a subagent | **0.9414** | 8m44s |

Both land above every baseline trial in the optimization run, which spanned 0.178 to 0.882 across ten scored trials.

This is **n=1 per arm**, and the README and `results/optimization-run.md` both say so rather than quoting the gap as an effect size. The task is bimodal: one configuration has produced 0.042, 0.905 and 0.909 across three runs. It is evidence that both changes reach the agent and do something, not a measured effect size. Raw results are committed at `results/candidates/measurement.csv`, and both candidate configs at `results/candidates/`.

Pushed as `bb6fa99`.

ericksoa (changes requested):
Thank you for the unusually candid run record. The revised documentation now distinguishes measured behavior from inference, and the exact-head focused suite passes locally (76 tests; SPDX clean). I still cannot approve this head because the two correctness boundaries from the prior review are not actually enforced.

1. `_assert_promotable_change_surface` runs inside the candidate-controlled `harbor_wrapper.py`. A candidate that edits the wrapper can remove or replace the assertion along with its scoring change. The new tests do not exercise rejection through a candidate import: `test_a_harness_edit_is_rejected` only asserts that the copied bytes differ, and the model-change test likewise compares normalized text. Please move this validation into code the candidate cannot edit, before the candidate wrapper is imported/executed, or have an external immutable launcher reject the candidate and prove that path with a malicious candidate regression.

2. Trace correlation still fails open. When more than one trace matches, `_await_trace` says the runs cannot be told apart but then selects the oldest and scores it. The new run hit this path three times. Manual inspection of those three cases does not make the mechanism reliable for future users. Please use a stable per-invocation identifier, distinct candidate agent names, or reject ambiguous trials.

One smaller exact-head issue: `git diff --check origin/main...HEAD` reports trailing whitespace on all five lines of `results/candidates/measurement.csv` (the file appears to carry CRLF endings).

The new out-of-loop measurements and the honest limitations are useful evidence, but they do not resolve these enforcement gaps.

phermosomore:
All three fixed. You were right on each, including that my tests were proving the wrong thing.

## 1. The pin was self-defeating

`_assert_promotable_change_surface` shipped inside `harbor_wrapper.py`, the file it protects. A candidate that rewrites the wrapper removes the check in the same edit. And `test_a_harness_edit_is_rejected` re-implemented the byte comparison inside the test rather than exercising a tampered candidate, so it asserted that bytes differ and nothing else. The one live check I ran appended a comment, which left the guard intact, so it raised and I concluded the mechanism worked.

`scorer/verify_candidates.py` is now the gate: outside `agent_source/`, never copied into a candidate, exits non-zero naming any candidate that changed the harness or a pinned `agent.yaml` block.

The regression you asked for is `test_a_candidate_that_deletes_the_guard_is_rejected`. It builds a candidate, **deletes the wrapper's own guard function and its call**, asserts the definition is gone, and asserts the external check still rejects it. `test_main_exits_nonzero_when_any_candidate_is_rejected` covers the CLI path.

I want to be straight about what this does not do. Harbor resolves the entry point inside the agent directory, so no in-loop check can be tamper-proof, and `verify_candidates.py` runs after trials rather than before: a violating candidate still consumes its trials before the run is rejected. The wrapper's check is now documented as fast feedback for accidental edits, not enforcement, and that residual gap is written into Known limitations rather than left implied.

## 2. Trace correlation no longer fails open

Each trial now invokes under a **unique agent name**, written into a temporary sibling config so the candidate's own `agent.yaml` is untouched and the pin still holds. `_await_trace` matches on that name alone and requires exactly one trace; anything else raises rather than scoring a guess.

The time window, the oldest-match tiebreak, and the warning that said runs could not be told apart and then scored one anyway are all gone, along with the now-dead `datetime` import.

## 3. CRLF

Confirmed on all five lines, from the Python `csv` writer. Converted to LF. My earlier check was `git diff --check` against `HEAD` rather than `origin/main...HEAD`, which is why I missed it.

## Verification on this head

- example suite 77 tests, up from 76
- `git diff --check origin/main...HEAD` clean
- catalog `--write` then `--check`: 26 examples across 7 browse groups
- license headers clean on 902 files
- `results/optimization-run.md` records where the pin is enforced and the timing gap, rather than repeating the claim that the wrapper enforces it

Pushed as `9c96415`.

ericksoa (changes requested):
Thank you for addressing the previous review directly. The unique per-trial agent name now gives trace correlation a real identity and `_await_trace` fails closed on multiple matches. The external malicious-wrapper regression also exercises the failure that the prior tests missed. On this exact head, the focused suite passes 83 tests, SPDX passes for 905 files, catalog metadata validates, and `git diff --check origin/main...HEAD` is clean.

I still cannot approve the current head because the expanded change introduces or leaves four correctness/safety gaps:

1. **The external verifier is not a complete change-surface boundary, and it runs after candidate code.** `verify_candidates.violations()` checks only `harbor_wrapper.py` and normalized `agent.yaml`. A candidate can change `pyproject.toml`, add an import-shadowing/helper module, or add/delete any other file outside `workspace/skills/`, and the verifier returns clean even though the README says everything except prompt/subagent/skills is pinned. More importantly, the candidate wrapper executes host-side before this manual post-run audit, so a rejected candidate has already run arbitrary modified harness/dependency code. Please make the immutable pre-execution boundary cover the whole candidate tree (allowlisting only the intended structured config fields and `workspace/skills/`, including rejecting unexpected files and symlinks), or restructure the immutable wrapper/dependencies outside `agent_source` so they are never candidate-controlled. Add regressions for modified `pyproject.toml` and an unexpected module/file. The residual post-execution timing gap is not just wasted trials; it is a host-code execution boundary.

2. **The refreshed public screenshots expose a local identity/path.** `assets/studio-trace.jpg` visibly contains `/Users/phermosomore/Downloads/AGENTIC_AI/nemoclaw-community/...`; `loop-overview.jpg` embeds the same capture. This contradicts the PR's release/compliance statement that no workspace paths are included. Please redact or replace both derived images and re-audit them at full resolution.

3. **The generated curl launcher does not JSON-escape its prompt.** `write_curl()` interpolates `${1}` inside a hand-written JSON string. A normal prompt containing a quote, backslash, or newline produces invalid JSON. Please construct the body with a JSON-aware tool and add a launcher test covering those characters.

4. **The generated dcode launcher silently reuses any existing global OpenAI provider block.** If `~/.deepagents/config.toml` already contains `[models.providers.openai]` for the real OpenAI endpoint or a different model list, `run.sh` skips configuration and then invokes `openai:<NeMo model>` against that unrelated provider. Please use an isolated/project-scoped configuration if supported, or validate/update the existing block without discarding unrelated settings, with a test for a pre-existing OpenAI provider.

The two original trace and CRLF findings are resolved, but approval has to cover the full current head, including the new export surface.

phermosomore:
All four addressed, and running a smoke test before replying caught a fifth that I had introduced. Head is `8d67a05`.

## 1. Change-surface boundary

You were right that it was not a boundary. It compared `harbor_wrapper.py` and a normalized `agent.yaml` and nothing else, so `pyproject.toml`, a new module, a deleted file or a symlink all verified clean.

Both the external verifier and the wrapper's own check now compare the **whole candidate tree** and allow exactly two openings: the `instructions` and `harnesses` blocks of `agent.yaml`, and anything under `workspace/skills/`. Symlinks are rejected outright, files the original does not ship are rejected rather than ignored, and deletions are caught.

Regressions cover a changed `pyproject.toml`, an unexpected module, a deleted file, a symlink, an added skill, and runtime output the agent and optimizer write inside the candidate.

**The timing gap is not closed, and the README now states it in your terms.** The candidate's wrapper executes host-side before any external audit, so this is a host-code execution boundary rather than wasted trials: by the time a violation is reported, modified harness code has already run. Closing it needs an entry point Harbor resolves outside the candidate, which `_scoped_import_path` does not offer, so the limitation says to run the loop where you would run any other code a model can edit. I would rather record that honestly than claim an enforcement I cannot deliver.

## 2. Screenshots

`studio-trace.jpg` carried an absolute workspace path including a username across three wrapped lines, and `loop-overview.jpg` embedded the same capture. Both are regenerated from a redacted source.

The region is blurred and then flattened to the panel background rather than blurred alone, since a Gaussian blur leaves recoverable residual and the redaction should not depend on how hard someone tries.

I re-audited the other three at full resolution: `studio-insight.jpg` carries only generated insight and experiment ids, `dcode-spoon.jpg` only FreeCAD's own panels, `mug-comparison.png` has no text. No file carries an identifying string in its bytes or an `Artist`, `Author`, `Comment` or `Software` metadata field. One limit worth stating: `tesseract` is not installed here, so that audit was visual inspection at full resolution rather than machine text extraction.

## 3. curl launcher

The body is built with `json.dumps` now. A test extracts the generated snippet, runs it with `He said "hi" \ and` plus a newline, and asserts the prompt round-trips. The previous form fails to parse on the first quote.

## 4. dcode launcher

Worse than reuse: I introduced the skip in the previous round while fixing a clobbering problem. It now sets `DEEPAGENTS_HOME` to a directory beside the project, so dcode's configuration is isolated and `~/.deepagents/config.toml` is neither read nor written. Verified against a live `dcode config path`, with a test pinning the isolation.

## 5. What the smoke test caught

The whole-tree check would have **rejected every candidate, baseline included, and failed every trial at import**. The optimizer writes `architecture.md` into each candidate after the change is final (`coder.py`, `fork.workdir / "architecture.md"`) to document the agent for the next round's proposer. It is not in the pristine `agent_source`, so the new check read it as an unexpected file.

Both halves skip it now, with the reason recorded beside the constant and a regression so the loop cannot be broken that way again.

## Verification on this head

A candidate copy carrying `architecture.md`, `artifacts/` and `traces/` was driven through the real wrapper path:

```text
guard passed, AGENT_DIR=agent-0, AGENT_NAME=cad-agent
trial config .trial-cad-agent-c6612827ca8d.yaml -> unique agent name cad-agent-c6612827ca8d
invoke rc = 0
trace correlated by unique name: 01a0d279-97dc-7131-b03d-d89cebbf7197
trial config removed
eval_iou span: 0.4423
```

Guard, per-trial config under a unique agent name, live invocation, single-match trace correlation, cleanup and scoring. A candidate that adds a module still refuses at import.

- example suite 92 tests, up from 83
- catalog metadata validates, 26 examples across 7 browse groups
- SPDX passes, 905 files
- `git diff --check origin/main...HEAD` clean

